PR #1, opened 2022-07-19
Take various updates from `fez` repository

cmeister2:
Here's the updates from `fez`. I suggest we rename `master` to `main` as well.

I'd love it if my pet feature made it in (fast RPM reading) but I've not included in this MR.

---

`fez` was a fork of `rpm-rs` with a number of maintenance fixes. This
commit takes those maintenance fixes and ports them into `rpm-rs`


Shinyzenith:
PR looks good apart from the failing tests :D tysm cmeister!!

Shinyzenith (dismissed):

Shinyzenith:
Also one minor nitpick: If the commit message could be a bit more descriptive it would help later on since not everyone knows about fez project

cmeister2:
> Also one minor nitpick: If the commit message could be a bit more descriptive it would help later on since not everyone knows about fez project

I've updated it with a little more detail.

Shinyzenith (dismissed):

drahnr (changes requested):
nit: there is a stray TODO that should be moved to an issue and be references, otherwise LGTM!

cmeister2:
> mit: there is a stray TODO that should be moved to an issue and be references, otherwise LGTM!

Fixed.

.github/dependabot.yml
```diff
@@ -0,0 +1,7 @@
+version: 2
+updates:
+- package-ecosystem: cargo
+  directory: "/"
+  schedule:
+    interval: daily
```
drahnr, line 6: ``
```suggestion
    interval: weekly
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,161 @@
+on:
+  push:
+    tags:
+      # Full version
+      - "[0-9]+.[0-9]+.[0-9]+"
```
drahnr, line 5: nit: prefixing with `v` might be a good idea to create a common prefix - it also seems to be a common pattern

.github/workflows/ci.yml
```diff
@@ -0,0 +1,161 @@
...
+      # Prerelease version
+      - "[0-9]+.[0-9]+.[0-9]+-*"
+
+  pull_request:
+    branches:
+      # Trigger on pull requests into main/master
+      - main
```
drahnr, line 12: I'll take care of the move to main after the merge

.github/workflows/ci.yml
```diff
@@ -0,0 +1,161 @@
...
+
+jobs:
+  check:
+    name: Check
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        rust:
+          - stable
+          - 1.57.0
```
drahnr, line 26: Does this align with the cargo manifest msrv? If not, that should be one a follow up PR, of which I can take care if

.github/workflows/ci.yml
```diff
@@ -0,0 +1,161 @@
...
+        with:
+          command: install
+          args: cargo-tarpaulin
+      - uses: actions-rs/cargo@v1
+        with:
+          command: tarpaulin
+          args: --out xml
+      # Commented out until a codecov token is added
+
+      # - uses: codecov/codecov-action@v3
```
drahnr, line 123: I'd prefer to delete commented code and introduce functional code all at once. We should move this to a separate PR. Do you want to take care of getting a token or should I?

src/rpm/signature/pgp.rs
```diff
@@ -31,7 +31,7 @@ impl traits::Signing<traits::algorithm::RSA> for Signer {
     /// it internally creates a copy until crate `pgp` provides
     /// a `Read` based implementation.
     fn sign<R: Read>(&self, data: R) -> Result<Self::Signature, RPMError> {
-        let passwd_fn = || String::new();
+        let passwd_fn = String::new;
```
drahnr, line 34: 👍

.github/workflows/ci.yml
cmeister2: Generally speaking I'm not a fan of a v prefix. Having the tag match the Cargo version (and also be semver) tends to be quite useful.

.github/workflows/ci.yml
cmeister2: It ought to - I'll double check when I'm back at a PC

.github/workflows/ci.yml
cmeister2: I don't mind. I can sort it out assuming I have access to do so.

Happy to delete commented-out code if you want - felt it was probably acceptable in the CI files temporarily.

.github/dependabot.yml
cmeister2: Don't mind.

.github/workflows/ci.yml
cmeister2: According to `cargo msrv` the actual MSRV is 1.56.1, so we should probably run with that.

.github/workflows/ci.yml
cmeister2: Have requested access for adding codecov. But I don't have access to add secrets to the repository.

drahnr: Added you. I'll grant access based on contributions over the next few weeks, so we have a few admins :)

.github/workflows/ci.yml
cmeister2: I've added in the v variants so we have a choice.

.github/workflows/ci.yml
drahnr: And also tie it down in the manifest

drahnr (dismissed):

.github/workflows/ci.yml
cmeister2: Ignore that, `cargo msrv` is blatantly lying 😂. I'll revert and make it 1.57.0 again.

```
`time v0.3.11` cannot be built because it requires rustc 1.57.0 or newer, while the currently active rustc version is 1.56.1
```

cmeister2:
@drahnr @Shinyzenith just need your reapproval whenever is convenient!

drahnr (approved):

Shinyzenith (approved):

PR #20, opened 2022-11-18
Fix packge sign issue

TommyLike:
1. Fix Seek&Read issue in SeqCursor
2. Fix md5 hash infinite loop when sign package
3. Add test for sign package and SeqCursor

Close #16

PR #21, opened 2022-11-18
Typo and Cleanup fix

TommyLike:
Add some typo fix as well as fix some clippy warning.

### DIFF
--- a/src/rpm/headers/header.rs
+++ b/src/rpm/headers/header.rs
@@ -538,7 +538,7 @@ impl Header<IndexTag> {
                         group: group.to_owned(),
                     },
                     mode: mode.into(),
-                    modified_at: utc.timestamp(mtime as i64, 0u32),
+                    modified_at: utc.timestamp_opt(mtime as i64, 0u32).unwrap(),
                     digest,
                     category: FileCategory::from_i32(flags).unwrap_or_default(),
                     size: size as usize,
--- a/src/rpm/headers/types.rs
+++ b/src/rpm/headers/types.rs
@@ -97,7 +97,7 @@ impl FileMode {
     }
 
     /// Usually this should be done with TryFrom, but since we already have a `From` implementation,
-    /// we run into this issue: `<https://github.com/rust-lang/rust/issues/50133>`
+    /// we run into this issue: <https://github.com/rust-lang/rust/issues/50133>
     pub fn try_from_raw(raw: i32) -> Result<Self, errors::RPMError> {
         let mode: FileMode = raw.into();
         mode.to_result()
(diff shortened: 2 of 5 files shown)


src/rpm/headers/header.rs
```diff
@@ -538,7 +538,7 @@ impl Header<IndexTag> {
                         group: group.to_owned(),
                     },
                     mode: mode.into(),
-                    modified_at: utc.timestamp(mtime as i64, 0u32),
+                    modified_at: utc.timestamp_opt(mtime as i64, 0u32).unwrap(),
```
dralley, line 541: e.g. why is this better

src/rpm/package.rs
```diff
@@ -110,8 +110,7 @@ impl RPMPackage {
         let rsa_signature_spanning_header_and_archive =
             signer.sign(&mut header_and_content_cursor)?;
 
-        // TODO FIXME verify this is the size we want, I don't think it is
```
dralley, line 113: This was verified I assume?

dralley:
Also let's run rustfmt (cargo fmt) so that the CI is completely clean

src/rpm/headers/header.rs
drahnr: `timestamp` got deprecated with the latest `chrono` release

src/rpm/package.rs
drahnr: I cross checked, this is indeed correct, I have a changeset ready once this an #20 are merged

src/rpm/headers/header.rs
drahnr: It will also make clippy pass again in other PRs once they are rebased.

TommyLike: yeah

TommyLike:
> @TommyLike Could you split the functional code change into a separate commit and use a descriptive commit message? This looks like more than just a typo fix.
There is no functional code change here.

src/rpm/package.rs
TommyLike: The code logic here is correct,remove the TODO FIXME tag here to avoid confusing, see: https://refspecs.linuxbase.org/LSB_3.1.1/LSB-Core-generic/LSB-Core-generic/pkgformat.html

src/rpm/headers/types.rs
```diff
@@ -97,7 +97,7 @@ impl FileMode {
     }
 
     /// Usually this should be done with TryFrom, but since we already have a `From` implementation,
-    /// we run into this issue: https://github.com/rust-lang/rust/issues/50133
+    /// we run into this issue: `<https://github.com/rust-lang/rust/issues/50133>`
```
drahnr, line 100:
```suggestion
    /// we run into this issue: <https://github.com/rust-lang/rust/issues/50133>
```

drahnr (approved):
Thank you!

dralley (approved):

Issue #4, opened 2022-07-22
Get publishing permissions on rpm-rs, or rename this crate


cmeister2:
We need to get publishing permissions on the rpm-rs crate, or rename this to something else, otherwise further changes are moot.

dralley:
No news.  The original contributor had some unrelated github activity 2 weeks ago but hasn't replied to emails or messages.

@cmeister2 I think, perhaps we should just continue publishing under the `fez` name for now.

cmeister2:
Couldn't we just rename to something like `rpm2` in the meantime, so we can actually get on with adding features? I'd be happy to use `fez` too, but there's a bit of hassle there with the feature I added.

dralley:
We could, I just figured if we already have a crate name, there's no need to create yet another one, that may eventually be abandoned anyway.

Is the feature not one that makes sense to upstream?

dralley:
@Richterrettich Trying again to reach you : /

@cmeister2 Do you feel it's a hard blocker?

olivierlemasle:
Hello @cmeister2 & @dralley,

Did you reach a decision regarding this crate? I'd be happy to contribute to this Richterrettich/rpm-rs fork, but I'd like to know if my pull requests are going to be merged some day, and if it a release of the forked crate will be made. If this fork is not more "maintained" that the original, that's a little bit sad...
Please understand that I feel entitled to nothing, that I do not blame you and that I really want to help :-)

@cmeister2 Regarding `fez`, **I'm not a lawyer**, but I'm not sure you had the right to change the license as you did. The StackExchange post you shared tells that you can use dependencies under one of them in a project under the other. It does not say you can re-license the work made by other authors. I do not think the part of 4 of the [Apache 2 license](https://apache.org/licenses/LICENSE-2.0.txt) enables you to change the license, even when you add a NOTICE file with the old license. This [other post from StackExchange](https://opensource.stackexchange.com/questions/9077/relicensing-from-apache-2-0-to-apache-or-mit) tells that you cannot re-license a project from Apache-2.0 to Apache-or-MIT without asking all the authors, so I guess it's the same thing (or worse!) to re-license from Apache-2.0 to MIT.

cmeister2:
@olivierlemasle looks like you're right; I'll have to fix that. Might be easiest to just blat the code here into that repo and forgo the change I made for now, or archive it.

[...]

PR #22, opened 2022-11-21
rename to just rpm

drahnr:
Follow up to #4 

I propose to craft a release as `0.0.1` under the new name `rpm` with the current state and iterate quickly and bump to `0.1.0` once a decent level of documentation exists.

Once merged, I'd also rename the name of this repo accordingly to `rpm`


dralley (approved):

drahnr:
Renamed repository to `rpm` :tada: - we should discuss when to slate the `0.0.1` release

dralley:
> I propose to craft a release as 0.0.1 under the new name rpm with the current state and iterate quickly and bump to 0.1.0 once a decent level of documentation exists.

That's fine with me, I'm slightly partial to picking up where the other crate left off in terms of versioning, but I don't feel strongly about it at all

drahnr:
We could do either, no strong opinion either.

dralley:
I don't think we need to overthink the "when" too much, the signing fix alone is justification for a release.

We need to update the changelog though - probably we should have been doing that in the PRs, but it slipped my mind.

PR #25, opened 2022-11-21
Add a few new  helpers to the Header struct for commonly used tags

dralley:


dralley:
Should we consider moving these helpers from `Header` to `RpmMetadata`, or even perhaps to `Package`?  It is unwieldy to have to `pkg.metadata.header`, and it fields a bit like a high-level / low-level layering violation.

src/rpm/headers/header.rs
```diff
@@ -237,6 +238,18 @@ where
             })
     }
 
+    pub(crate) fn get_entry_number_data(&self, tag: T) -> Result<i64, RPMError> {
+        let entry = self.find_entry_or_err(&tag)?;
+        entry
+            .data
+            .as_number()
+            .ok_or_else(|| RPMError::UnexpectedTagDataType {
+                expected_data_type: "numeric",
```
dralley, line 247: No such type exists officially, but I'm not sure how else to express that it could be any numeric type

src/rpm/headers/header.rs
```diff
@@ -441,8 +445,77 @@ impl Header<IndexTag> {
...
+    // }
+
+    #[inline]
+    pub fn get_packager(&self) -> Result<&str, RPMError> {
+        self.get_entry_string_data(IndexTag::RPMTAG_PACKAGER)
+    }
+
+    #[inline]
+    pub fn get_build_time(&self) -> Result<i64, RPMError> {
+        // Typically this is stored as an i32, however, exposing it as such is a bad idea.
```
dralley, line 485: Annoyingly, `rpm` seems to store this as `i32` but then they claim that it's year 2038 safe because they treat it like a u32 everywhere.  I'm sure that's true, but it's a mess.  Let's not do that.

src/rpm/headers/header.rs
```diff
@@ -401,18 +414,9 @@ impl Header<IndexSignatureTag> {
...
-        self.get_entry_string_data(IndexTag::RPMTAG_PAYLOADFORMAT)
-    }
-
-    #[inline]
-    pub fn get_payload_compressor(&self) -> Result<&str, RPMError> {
-        self.get_entry_string_data(IndexTag::RPMTAG_PAYLOADCOMPRESSOR)
-    }
-
-    #[inline]
-    pub fn get_file_checksums(&self) -> Result<&[String], RPMError> {
```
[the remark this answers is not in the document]
dralley: They aren't, just relocated further down

dralley:
@drahnr What do you think about the questions in the first comment?

src/rpm/headers/header.rs
drahnr: Can we just make it an int64 without causing any further incompatibilites?

src/rpm/headers/header.rs
drahnr: I think it boils down to, if it's allowed to change the number representation or if they're fixed per tag as given by the specification (well, if there was one..). So I think if it works with the defacto impl a numeric is an abstraction over all int types so this is fine, it's just an error right?

src/rpm/headers/header.rs
dralley: Do you mean the value of the IndexData itself?

dralley: `rpm` may in fact treat *all* integers as unsigned despite the name https://github.com/rpm-software-management/rpm/blob/master/docs/manual/tags.md#scalar-types

drahnr: My largest concern is if existing implementations assume fixed types per tag, such that other implementations would read only individual bytes of i.e. a larger integer.

src/rpm/headers/header.rs
dralley: FWIW, librpm has this, the function is `rpmtdGetNumber` and it just returns a uint64 regardless of what the tag type was.

drahnr: Interesting. If that's used consistently within rpm, we could technically make the serialized version whatever type we'd want to.

src/rpm/headers/header.rs
dralley: I've asked for some clarification on how the types are supposed to work, because there's a lot of inconsistency.  The docs claim the integers are unsigned but the signatures and types suggest otherwise.

src/rpm/headers/header.rs
drahnr: Let's call it `"scalar"` for the sake of consistency

drahnr (approved):

src/rpm/headers/header.rs
dralley: I've filed https://github.com/rpm-rs/rpm/issues/41

> <_ffesti_> and yes all integer tags are unsigned

drahnr: I assume IRC? We should document these impl details..

dralley:
@drahnr @cmeister2 Do you think the change added in the last commit makes sense?  Motivation [here](https://github.com/rpm-rs/rpm/pull/25#issuecomment-1324315419)

It might even make sense to move up to `Package` itself, I'm not sure which I prefer yet.

src/rpm/headers/header.rs
```diff
@@ -383,16 +382,6 @@ impl Header<IndexSignatureTag> {
         Ok(())
     }
 
-    #[inline]
-    pub fn get_file_ima_signatures(&self) -> Result<&[String], RPMError> {
-        self.get_entry_data_as_string_array(IndexSignatureTag::RPMSIGTAG_FILESIGNATURES)
-    }
-
-    #[inline]
-    pub fn get_file_ima_signature_length(&self) -> Result<u32, RPMError> {
```
dralley, line 392: Sidenote, this kinda feels like an internal detail, when exposed to the user Rust types ought to know their own lengths.  I haven't looked into it deeply though - does that sound right?

drahnr: Sounds good to me

dralley: Are we 100% certain the signatures are not binary and should never contain a null character except as termination?

dralley: https://github.com/rpm-software-management/rpm/blob/master/docs/manual/tags.md#optional-file-information

Claims they are hex-encoded, so I suppose it should be fine.

drahnr: Yeah, sure of that.

drahnr (approved):

PR #42, opened 2022-11-29
Use unsigned integers instead of signed integers for handling tag data

dralley:


src/rpm/headers/header.rs
```diff
@@ -191,67 +191,67 @@ where
...
-    pub(crate) fn get_entry_i16_array_data(&self, tag: T) -> Result<Vec<i16>, RPMError> {
+    pub(crate) fn get_entry_data_as_u16_array(&self, tag: T) -> Result<Vec<u16>, RPMError> {
         let entry = self.find_entry_or_err(&tag)?;
         entry
             .data
-            .as_i16_array()
+            .as_u16_array()
             .ok_or_else(|| RPMError::UnexpectedTagDataType {
-                expected_data_type: "i64 array",
+                expected_data_type: "uint64 array",
```
dralley, line 200: I'm conflicted about tweaking the names from the documented ones even if it makes it clearer.  Thoughts?

src/rpm/builder.rs
```diff
@@ -151,7 +152,9 @@ impl RPMBuilder {
                 .modified()?
                 .duration_since(UNIX_EPOCH)
                 .expect("something really wrong with your time")
-                .as_secs() as i32,
+                .as_secs()
+                .try_into()
+                .unwrap(),
```
cmeister2, line 157: This goes for every instance - should this be handled (by conversion into an RPMError) or expect()ed ? Or are we assuming that this is completely infallible (what happens if we parse a file with a modified date in 2040 for example - or is this just a limitation of the RPM spec?)

dralley: Well realistically for that purpose RPM needs to stop storing the tag as a 32 bit integer.  If all the integers are unsigned then that means it should be OK for any date from 1970 out to about 2100 so there is at least no "year 2038 problem"

https://github.com/rpm-software-management/rpm/issues/1084

dralley:
@drahnr How does this look?

src/rpm/headers/types.rs
```diff
@@ -3,12 +3,12 @@ use crate::{constants::*, errors};
...
 pub struct RPMFileEntry {
-    pub(crate) size: i32,
+    pub(crate) size: u32,
     pub(crate) mode: FileMode,
-    pub(crate) modified_at: i32,
+    pub(crate) modified_at: u32,
     pub(crate) sha_checksum: String,
     pub(crate) link: String,
-    pub(crate) flag: i32,
+    pub(crate) flag: u32,
```
dralley, line 11: https://github.com/rpm-rs/rpm/issues/52

src/rpm/builder.rs
```diff
@@ -439,20 +444,20 @@ impl RPMBuilder {
...
             let index = self
                 .directories
                 .iter()
                 .position(|d| d == &entry.dir)
                 .unwrap();
-            dir_indixes.push(index as i32);
+            dir_indixes.push(index as u32);
             base_names.push(entry.base_name.to_owned());
-            file_verify_flags.push(-1);
+            file_verify_flags.push(u32::MAX); // TODO: convert to bitfield or something so that we can use actual flag semantics
```
dralley, line 456: https://github.com/rpm-rs/rpm/issues/52

Cargo.toml
```diff
@@ -41,8 +41,6 @@ log = "0.4"
 itertools = "0.10"
 hex = { version = "0.4", features = ["std"] }
 zstd = "0.12.0"
-# Workaround for https://github.com/gyscos/zstd-rs/issues/177
-zstd-sys = "=2.0.4+zstd.1.5.2"
```
drahnr, line 45: Should not be part of this PR

src/constants.rs
```diff
@@ -5,6 +5,8 @@
 
 use std::fmt::Display;
 
+// TODO: these constants should probably be usize, `rpm` uses signed integers in a lot of places that
+// it probably should not.
```
drahnr, line 9: I'd prefer to reference a gh issue within TODOs

src/rpm/builder.rs
```diff
@@ -1,4 +1,5 @@
 use std::collections::{BTreeMap, BTreeSet};
+use std::convert::TryInto;
```
drahnr, line 2: OOS: Should we consider moving to edition 2021? https://doc.rust-lang.org/core/prelude/rust_2021/index.html

src/rpm/builder.rs
drahnr: Generally speaking, an `RPMError`. In this particular case, I think a `expect()` should suffice.

src/rpm/builder.rs
```diff
@@ -307,7 +312,7 @@ impl RPMBuilder {
                     header_digest_sha1.as_str(),
                     header_and_content_digest_md5.as_slice(),
                 )
-                .build(header_and_content_len as i32)
+                .build(header_and_content_len.try_into().unwrap())
```
drahnr, line 315: Could you explain why we need to replace the `as` cast? The only advantage I see is a actionable panic rather than a incorrect value, which is great! But then why are we not replacing all of them? https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=c7f6783c4f4b1eb6709102a1f54c7044

src/rpm/builder.rs
drahnr: Could you reference the issue within the TODO comment?

drahnr (changes requested):
Generally the core change looks good :+1: ,  but it seems there is at least one change that's unrelated/accidental, plus a few nits that ought to be resolved before merging.

src/rpm/builder.rs
dralley: >actionable panic rather than a incorrect value, which is great! 

That's basically it

>But then why are we not replacing all of them?

I mostly just changed the ones I needed to for this PR, but I think the majority of other instances of casting relate to the constants which are currently `isize`, but they're all manually defined and small numbers so they shouldn't be problematic.

dralley:
@drahnr How about now

src/constants.rs
```diff
@@ -5,6 +5,7 @@
 
 use std::fmt::Display;
 
+// @todo: https://github.com/rpm-rs/rpm/issues/54
```
drahnr, line 8:
```suggestion
// @todo: <https://github.com/rpm-rs/rpm/issues/54>
```

src/rpm/headers/header.rs
```diff
@@ -575,7 +575,7 @@ pub struct FileOwnership {
 }
 
 /// Declaration what category this file belongs to
-/// @todo must be bitflags
+/// @todo must be bitflags: https://github.com/rpm-rs/rpm/issues/52
```
drahnr, line 578:
```suggestion
/// @todo must be bitflags: <https://github.com/rpm-rs/rpm/issues/52>
```

src/rpm/headers/types.rs
```diff
@@ -3,12 +3,12 @@ use crate::{constants::*, errors};
...
 pub struct RPMFileEntry {
-    pub(crate) size: i32,
+    pub(crate) size: u32,
     pub(crate) mode: FileMode,
-    pub(crate) modified_at: i32,
+    pub(crate) modified_at: u32,
     pub(crate) sha_checksum: String,
     pub(crate) link: String,
-    pub(crate) flag: i32,
+    pub(crate) flag: u32, // @todo: https://github.com/rpm-rs/rpm/issues/52
```
drahnr, line 11:
```suggestion
    pub(crate) flag: u32, // @todo: <https://github.com/rpm-rs/rpm/issues/52>
```

src/rpm/headers/types.rs
```diff
@@ -35,6 +35,7 @@ const PERMISSIONS_BIT_MASK: u16 = 0o7777; // bit representation = "0000111111111
 const REGULAR_FILE_TYPE: u16 = 0o100000; //  bit representation = "1000000000000000"
 const DIR_FILE_TYPE: u16 = 0o040000; //      bit representation = "0100000000000000"
 
+// @todo: https://github.com/rpm-rs/rpm/issues/52
```
drahnr, line 38:
```suggestion
// @todo: <https://github.com/rpm-rs/rpm/issues/52>
```

drahnr (approved):
Looks good, the only thing that's missing is documenting the implementation detail deviation from upstream.

I think an additional file in the style of an ADR as a markdown file that can be referenced would make sense, but that's for another PR.

src/rpm/headers/types.rs
dralley: Why?

dralley:
>Looks good, the only thing that's missing is documenting the implementation detail deviation from upstream.

Which deviation?  It seems like `rpm` just stores untyped bytes and casts them at time of use based on the tag, but it uses unsigned ints when doing so.  Or do you mean the names of the types?

drahnr:
I mean the different implementation details of treating them as unsigned where as treating them as signed upstream at the initial parsing state. If I was to look at the code, that'd be so erging very apparent and should be documented in some way.

dralley:
Are you referring to https://github.com/rpm-software-management/rpm/blob/master/lib/header.c#L760-L789?

If you look at the definition of `htonl` and `htonll`, it's using unsigned integers.  The annoying thing is just how inconsistent they are about what types are used where, casting back and forth constantly.

src/rpm/builder.rs
drahnr: I'd prefer if this was done all in one go, or it's only done for the particular item, but iiuc it's neither? Sorry for being picky, I simply want to avoid this project going the same path as upstream.

drahnr:
What's left to do:
* [ ] document why we're changing this, it's non obvious why there is a deviation from the C rpm source-  yes,  "original C rpm" makes no sense and is a mess, but _we_ should strive for clarity _why_ we make certain choices. The documentation is still missing. One way of documenting would be `Architecture Decision Records` - just a few markdown files in a specified folder covering the design constraints and ultimate the rationale for the choice. I am open for alternatives, but there _must_ be documentation
* [ ] make sure the changeset is applied thoroughly to the entire codebase

dralley:
@drahnr I think I remember seeing that you had added some commits but I believe I force-pushed over them this morning by accident.

In terms of documentation, honestly I'd rather just try to complain upstream and get it fixed (or at least documented) there `https://github.com/rpm-software-management/rpm/pull/2316`

drahnr:
Ok, I see in that case it should be fine

Regarding the commits, it was only about adding `<` `>` around the links ac Ording to markdown spec

dralley:
I'm confused why Markdown formatting is desirable when they will never be rendered as such - most are just standard comments, and the two that are doc comments, are on private functions.  I would think it would just get in the way when looking at the source code and trying to copy the link.

src/rpm/headers/header.rs
```diff
@@ -575,7 +575,6 @@ pub struct FileOwnership {
 }
 
 /// Declaration what category this file belongs to
-/// @todo must be bitflags
```
dralley, line 578: I'm not sure that using bitflags in this particular spot makes sense, since AFAIK the file can only belong to one category.

drahnr:
> I'm confused why Markdown formatting is desirable when they will never be rendered as such - most are just standard comments, and the one that is part of a doc comment, is on a private function. I would think it would just get in the way when looking at the source code and trying to copy the link - source code readers being the intended audience for any "todo" comment.

Because if your editor is markdown aware, you don't need to copy it, but just click / toggle to the link and hit enter :) but that might be just me

drahnr:
@dralley do you intend to get rid of _all_ `i32` instances, since to my understanding none of the ones left bear any meaning for negative numbers, and there are quite a few. I am reluctant to merge a partial changeset. I'd prefer if we'd use proper link delimiters, such that tooling like i.e. `cargo-spellcheck` can be used on _all_ comments, dev, public, private whatever.

Other than that, LGTM

The remaining files with `i32`s where there is no use for negative numbers: 
```
src/constants.rs
src/errors.rs
src/rpm/builder.rs
src/rpm/headers/signature_builder.rs
src/rpm/headers/header.rs
src/rpm/headers/types.rs
```

dralley:
>     Because if your editor is markdown aware, you don't need to copy it, but just click / toggle to the link and hit enter :) but that might be just me

My editor (VSCode) detects links even without the angle brackets so I've never noticed a difference.  If you feel strongly though, I can add them.  They at least don't seem to interfere with VSCode.

>@dralley do you intend to get rid of all i32 instances, since to my understanding none of the ones left bear any meaning for negative numbers, and there are quite a few. 

For the moment it's just limited to integers involved in handling tag data, specifically.  I filed a separate issue to update the constants and any other places I would consider out of scope for the PR for now.

drahnr:
In my opinion, that is insufficient. It's in another repo, in a closed issue/PR, hard to discover!
What I think is the reasonable bare minimum to track these implementation changes as content changes.
Just a text file, describe the motivation, link it from the most pivotal point in code, add the links to issues and issue comments to that file, but also provide sufficient rationale right there to understand the why.

I don't see any downside to the outlined approach, and I do feel strongly about it.

Other than that, it's good to go. But that rationale has to be written down, it's going to be a change through the entirety of the codebase at the tag level.

drahnr:
> > ```
> > Because if your editor is markdown aware, you don't need to copy it, but just click / toggle to the link and hit enter :) but that might be just me
> > ```
> 
> My editor (VSCode) detects links even without the angle brackets so I've never noticed a difference, but they don't seem to interfere with VSCode so I will add them back.

Interesting, my fault, I really thought it only detected the marked urls in markdown, maybe that's true for `http` prefixed ones. It still doesn't hurt to add them, they're spec'd out <https://www.markdownguide.org/basic-syntax#urls-and-email-addresses>, and other tools, pick them up better than inline unmarked links

drahnr (approved):

PR #55, opened 2022-12-07
Remove workaround for zstd package issue, update deps

dralley:


drahnr (approved):

dralley:
@drahnr I believe there are "required checks" for checks that no longer exist, but I can't seem  to find where those are configured

dralley:
Never mind, I found it.

dralley:
@drahnr Do we really need clippy and rustfmt running with Rust 1.59?  The outputs there shouldn't really change, the important thing is that it builds and passes tests.

drahnr:
I don't recall why I added them, so feel free to remove after checking the commit that added them.

PR #56, opened 2022-12-07
Delete unused / unecessary Github CI steps

dralley:


.github/workflows/ci.yml
```diff
@@ -144,18 +141,3 @@ jobs:
         with:
           command: tarpaulin
           args: --out xml
-
-  publish:
```
dralley, line 148: We'll probably want to bring this back, but probably in a separate action triggered by a tag push / release creation.

As it is all the publish logic is missing anyway.

cmeister2:
I don't think these steps are unnecessary - stable is not necessarily the same as the MSRV. The point of the fixed version is to be at the MSRV.

Is it also worth deleting the publish logic if we think we're going to go that way (publish on tag)? The actual publish logic should be in the git history...

dralley:
>I don't think these steps are unnecessary - stable is not necessarily the same as the MSRV. The point of the fixed version is to be at the MSRV.

We're still compiling and running tests on the MSRV.  Running clippy, rustfmt, and an additional `cargo check` there is unnecessary though.  I can't see that providing any benefit.

>Is it also worth deleting the publish logic if we think we're going to go that way (publish on tag)? The actual publish logic should be in the git history...

There isn't any publish logic here, just a step named "publish" that doesn't do anything.  When we figure out what we want to do in that respect, we can add it all in one commit.

cmeister2 (approved):
LGTM

drahnr:
Now I do remember and I think we need to check against msrv declared in the manifest.

Building might be sufficient, but there must be a automated check

drahnr (changes requested):
We need a way to assure msrv compliance us not broken accidentally, this PR would remove that.

dralley:
`check` is just a strict subset of the compile process though?

drahnr:
Yes, `build` for msrv is sufficient

dralley:
This PR leaves that intact.  https://github.com/rpm-rs/rpm/pull/56/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fL47

I'm confused whether anything still needs to be changed.

drahnr:
@dralley I think it can be merged if we re-add the `- 1.60.0` to the 'non default feature set' task matrix. Otherwise only check the default features to be semver compliant. (I am aware that this is not a full test which would require a lot more combinations of features, but it's better than nothing).

Sorry for the back and forth

dralley:
> @dralley I think it can be merged if we re-add the - 1.60.0 to the 'non default feature set' task matrix.

That matrix is for "no default features" not "non default features", it's a more limited set of features than the set which is currently being tested which includes all of the default features.

.github/workflows/ci.yml
```diff
@@ -63,6 +62,7 @@ jobs:
       matrix:
         rust:
           - stable
+          - 1.60.0
```
dralley, line 65: Note that we weren't actually running this before

drahnr: Yeah, we should have. I think the gha generally need some more attention. But that's out of scope

drahnr (approved):

PR #62, opened 2022-12-19
Remove get_file_ima_signature_length()

dralley:
Remove the helper function `get_file_ima_signature_length` after a discussion with RPM developers - the tag should be considered obsolete and therefore we ought not to encourage its use with a helper.

https://github.com/rpm-software-management/rpm/commit/b7e9a60370d910febbe8c69e2d2cac7512051eb3

- 🦣 Legacy
- 🪣 Misc

### DIFF
--- a/src/rpm/package.rs
+++ b/src/rpm/package.rs
@@ -223,18 +223,6 @@ impl RPMPackageMetadata {
     }
 
     #[inline]
-    pub fn get_file_ima_signatures(&self) -> Result<&[String], RPMError> {
-        self.signature
-            .get_entry_data_as_string_array(IndexSignatureTag::RPMSIGTAG_FILESIGNATURES)
-    }
-
-    #[inline]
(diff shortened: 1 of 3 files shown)


src/rpm/package.rs
```diff
@@ -335,6 +323,12 @@ impl RPMPackageMetadata {
             .get_entry_data_as_string_array(IndexTag::RPMTAG_FILEDIGESTS)
     }
 
+    #[inline]
+    pub fn get_file_ima_signatures(&self) -> Result<&[String], RPMError> {
+        self.signature
+            .get_entry_data_as_string_array(IndexSignatureTag::RPMSIGTAG_FILESIGNATURES)
+    }
```
dralley, line 330: This is just moved, if you look at the functions above and below it makes more sense down here.

drahnr (approved):

PR #64, opened 2022-12-28
Inspect uses of unwrap() and expect() and improve them where possible

### LINKED ISSUE #46: Eliminate as many uses of unwrap() and expect() and other sources of panics as possible
We should be returning errors to the user to the greatest extent possible rather than crashing their programs outright.  `unwrap()` and `expect()` are useful for speeding up development but they're not great practice for a library long-term.

 In some places it may be unavoidable or the best option, I'm not convinced we need to ban them outright, but it would be a good exercise to take a look at where we're using them and fix it up.  I definitely see a few that could be improved.

dralley:
closes #46

Eliminates a few potential sources of panic!(), improve message in cases where it is left as-is.

- 🪣 Misc

### DIFF
--- a/src/rpm/builder.rs
+++ b/src/rpm/builder.rs
@@ -160,7 +160,7 @@ impl RPMBuilder {
             .await?
             .modified()?
             .duration_since(UNIX_EPOCH)
-            .expect("something really wrong with your time")
+            .expect("system time predates the Unix epoch?")
             .as_secs();
 
         self.with_file_async_inner(input.compat(), modified_at, options)
@@ -183,7 +183,7 @@ impl RPMBuilder {
             .await?
             .modified()?
             .duration_since(UNIX_EPOCH)
-            .expect("something really wrong with your time")
+            .expect("system time predates the Unix epoch?")
             .as_secs();
 
         self.with_file_async_inner(input, modified_at, options)
@@ -227,7 +227,7 @@ impl RPMBuilder {
                 .metadata()?
                 .modified()?
                 .duration_since(UNIX_EPOCH)
-                .expect("something really wrong with your time")
+                .expect("system time predates the Unix epoch?")
                 .as_secs()
                 .try_into()
                 .expect("file mtime is wrong, too large to be stored as uint32"),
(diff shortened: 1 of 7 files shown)


src/rpm/builder.rs
```diff
@@ -277,9 +279,10 @@ impl RPMBuilder {
         let hash_result = hasher.finalize();
         let sha_checksum = hex::encode(hash_result); // encode as string
         let entry = RPMFileEntry {
+            // file_name() should never fail because we've checked the special cases already
             base_name: pb.file_name().unwrap().to_string_lossy().to_string(),
             size: content.len() as u32,
-            content: Some(content),
```
dralley, line 282: It doesn't appear that we're ever using the `None` value, we don't create any `RPMFileEntry` with default values that I can tell, nor do any functoins construct one without passing in a `content`.

There might be some reason for this I am unaware of, though.

src/rpm/headers/lead.rs
```diff
@@ -94,17 +94,16 @@ impl Lead {
         let mut name_arr: [u8; 66] = [0; 66];
         name_arr.copy_from_slice(name);
 
-        //save unwrap here since we've checked length of slices.
         Ok(Lead {
-            magic: magic.try_into().unwrap(),
+            magic: RPM_MAGIC,
```
dralley, line 98: We've already checked for equality with the constant value, so why not just use the constant value.

src/rpm/builder.rs
```diff
@@ -361,7 +364,11 @@ impl RPMBuilder {
                     header_digest_sha1.as_str(),
                     header_and_content_digest_md5.as_slice(),
                 )
-                .build(header_and_content_len.try_into().unwrap())
+                .build(
+                    header_and_content_len
+                        .try_into()
+                        .expect("signature header + signature length must be <4gb"),
```
dralley, line 370: We can do away with these once https://github.com/rpm-rs/rpm/issues/65 is handled.

src/rpm/builder.rs
```diff
@@ -260,9 +260,11 @@ impl RPMBuilder {
                 path: dest.clone(),
                 desc: "no parent directory found",
             })?;
+
         let (cpio_path, dir) = if dest.starts_with('.') {
```
drahnr, line 264: Using https://doc.rust-lang.org/std/primitive.str.html#method.strip_prefix could avoid the `unwrap` in line `267` entirely

drahnr (approved):

PR #69, opened 2023-01-01
Use proper flag semantics instead of passing around raw integers

### LINKED ISSUE #52: Use proper flag semantics instead of passing around integers
There are a couple of places where raw integers are passed around as "flags", where individual bits of would be considered to set various flags.  Most of this functionality behind those flags is missing, but in any case we ought to evaluate how to most ergonomically expose it, e.g. using something like the `bitflags` crate instead of plain integers.

dralley:
closes #52

### DIFF
--- a/src/constants.rs
+++ b/src/constants.rs
@@ -5,6 +5,8 @@
 
 use std::fmt::Display;
 
+use bitflags::bitflags;
+
 pub const HEADER_IMAGE: u32 = 61;
 pub const HEADER_SIGNATURES: u32 = 62;
 pub const HEADER_IMMUTABLE: u32 = 63;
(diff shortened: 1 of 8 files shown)


dralley:
@drahnr I think that this line might logically unsound, because these flags are not necessarily mutually exclusive.  I'm not 100% sure.  Thoughts?

https://github.com/rpm-rs/rpm/blob/b015c277251727c57a9f662ae953b8fad077954a/src/rpm/package.rs#L455

src/rpm/package.rs
```diff
@@ -819,7 +821,7 @@ impl RPMPackageMetadata {
                             mode: mode.into(),
                             modified_at: utc.timestamp_opt(mtime as i64, 0u32).unwrap(), // shouldn't fail as we are using 0 nanoseconds
                             digest,
-                            category: FileCategory::from_u32(flags).unwrap_or_default(),
+                            flags: FileFlags::from_bits(flags).expect("Unexpected rpm file flags found"),
```
drahnr, line 824: Should be an error, since there it's within a `try_fold` anyways, `panic` seems unnecessary.

drahnr (changes requested):
I like it, but we should avoid any panic-ing behavior in parsing packages, regardless if incoherent or not.

dralley:
FWIW, while it's still drafted I don't necessarily expect a review, still making some changes

src/constants.rs
```diff
@@ -472,57 +474,102 @@ pub const RPM_MAGIC: [u8; 4] = [0xed, 0xab, 0xee, 0xdb];
...
+        const PUBKEY = 1 << 11;	// %%pubkey
+        const ARTIFACT	= 1 << 12;	// %%artifact
+    }
+}
+
+// ????
+// pub const RPMFILE_PARSE_EXCLUDE: u32	= (1 << 16),	// %%exclude
+// pub const RPMFILE_PARSE_DOCDIR: u32	= (1 << 17),	// from %%docdir
+// pub const RPMFILE_PARSE_DIR: u32		= (1 << 18),	// from %%dir
+// pub const RPMFILE_PARSE_SPECIALDIR: u32	= (1 << 19),	// special %%doc
```
dralley, line 563: There's still some gaps missing but we can deal with it later.  I have no idea what these are actually for.

src/rpm/package.rs
```diff
@@ -699,23 +701,23 @@ impl RPMPackageMetadata {
...
             })
     }
 
     /// Extract a the set of contained file names including the additional metadata.
     pub fn get_file_entries(&self) -> Result<Vec<FileEntry>, RPMError> {
         // rpm does not encode it, if it is the default md5
-        let algorithm = self.get_file_digest_algorithm().unwrap_or_default();
+        let algorithm = self
+            .get_file_digest_algorithm()
+            .unwrap_or(DigestAlgorithm::Md5);
```
dralley, line 720: Change: deleted `Default` impl for DigestAlgorithm as it should probably be contextual

Change: renamed a few fields named after `sense` to `flags`

PR #70, opened 2023-01-07
Fix compat tests + run in CI

### LINKED ISSUE #43: `test-with-podman` tests don't run in CI
I just tried running `test-with-podman` tests locally and they didn't compile. I don't think the CI is currently set up to do the testing either - it probably should.

olivierlemasle:
# PR Content Desc

- 🩹 Bug Fix

Closes #43

## Changes proposed by this PR

- Fix compatibility tests
- Run compatibility tests in CI

## 📜 Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,6 +84,25 @@ jobs:
           command: test
           args: ${{ matrix.flags }}
 
+  compat-tests:
+    name: Compatibility Test Suite
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        rust:
+          - stable
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: ${{ matrix.rust }}
+          override: true
+      - uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: --features test-with-podman
+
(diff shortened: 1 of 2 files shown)


cmeister2 (commented):
Is this test being run? I'm surprised that podman and yum commands would work on ubuntu-latest

olivierlemasle:
`podman` is actually installed in GitHub ubuntu runner image: https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#tools (I was surprised too!)

`rpm` is not needed for compat tests (`rpm`, `yum` and `dnf` are  used **inside** podman containers).

See the logs here: https://github.com/rpm-rs/rpm/actions/runs/3863408316/jobs/6585580273#step:4:419

dralley:
Couldn't we just have the CI run inside a Fedora container instead of manually spawning podman from inside the tests?  That seems simpler.

Like [this](https://github.com/rpm-software-management/createrepo_c/blob/master/.github/workflows/ci-python.yml#L9)

olivierlemasle:
I think the main advantages of the current solution are:
- tests are run on muliple OSes (currently `fedora:36`, `centos:stream9` and `centos:7`)
- the same integration tests can be executed easily both outside CI (just requires `cargo` and `podman`) and in CI

dralley (approved):

PR #83, opened 2023-02-26
Add package build time

### LINKED ISSUE #82: Add package build timestamp

troelsarvin:
- 🦚 Feature

Closes #82


troelsarvin:
Oops, should have handled test properly, regretting this PR.

PR #84, opened 2023-02-26
Adding build timestamp

### LINKED ISSUE #82: Add package build timestamp

troelsarvin:
- 🦚 Feature

Closes #82, passes "cargo test".

### DIFF
--- a/src/rpm/package.rs
+++ b/src/rpm/package.rs
@@ -240,6 +240,12 @@ impl RPMPackageMetadata {
     }
 
     #[inline]
+    pub fn get_buildtime(&self) -> Result<u32, RPMError> {
+        self.header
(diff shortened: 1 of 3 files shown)


src/rpm/package.rs
```diff
@@ -239,6 +239,12 @@ impl RPMPackageMetadata {
         self.header.get_entry_data_as_u32(IndexTag::RPMTAG_EPOCH)
     }
 
+    #[inline]
+    pub fn get_buildtime(&self) -> Result<u32, RPMError> {
```
dralley, line 243: See line 297, this already exists

troelsarvin:
Another (smaller) pull request will be brought forward making use of the existing get_build_time.

PR #85, opened 2023-02-26
Put build time into generated RPMs

### LINKED ISSUE #82: Add package build timestamp

troelsarvin:
- 🦚 Feature

Closes #82

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,7 +43,7 @@ jobs:
       matrix:
         rust:
           - stable
-          - 1.60.0
+          - 1.63.0
     steps:
       - uses: actions/checkout@v3
       - uses: actions-rs/toolchain@v1
@@ -62,7 +62,7 @@ jobs:
       matrix:
         rust:
           - stable
-          - 1.60.0
+          - 1.63.0
         flags:
           - "--all-features"
           - "--no-default-features"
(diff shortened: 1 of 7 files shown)


dralley:
@troelsarvin There's no need to create new PRs, you can fix up the code in the existing ones :)  If you don't know how I can help.

Could you please squash the latter two commits into the first?

troelsarvin:
"cargo fmt" has been run now. However, I don't know how to squash commits when I have already pushed.

dralley:
Here is a tutorial on doing it from the command line but I am sure if you are using a GUI there will be a way to do the same thing.

https://thoughtbot.com/blog/git-interactive-rebase-squash-amend-rewriting-history

troelsarvin:
Thanks for the links. I found they didn't completely fit the situation where I need to rebase already-pushed commits. Now I've force-pushed the four separate commits into one. I hope that can work. Otherwise, I'll resort to sending a new, separate merge request.

dralley:
Force pushing to your own branch is perfectly fine and is exactly what you have to do to rewrite history on a PR, so you did it correctly.

dralley (approved):

dralley:
@troelsarvin Could you rebase on top of master branch?  It will fix the issue causing the CI to fail, which is unrelated to your changes.

* `git fetch $upstream-repo-remote-name` (or checkout main and git pull from the upstream)
* `git rebase master`
* `git push -f`

dralley:
So that is technically a merge not a rebase :)  What you want to do is fetch the upstream remote, then `git rebase $upstream-remote-name`, then force-push

troelsarvin:
I hope it's useful now.

dralley (approved):

PR #91, opened 2023-03-05
Support files >4gb, default to 64bit tag sizes for tag-related data

### LINKED ISSUE #65: Support packages and files with sizes larger than u32::MAX
RPM uses 64 bit tags when a file size goes beyond the maximum storable in a 32 bit unsigned integer.  See: https://rpm.org/devel_doc/large_files.html

It appears we are handling this case while parsing, but our API doesn't allow creating such packages, so the API needs to change to accommodate and we will need to handle those tags properly when writing RPMs.

dralley:
Use 64bit file tags to support file sizes >4gb.  All versions of `rpm` starting with 4.6.0 (circa 2008) support and look for the 64bit tags by default so there is little reason to support the fallback logic.

- 🩹 Bug Fix
- 🦚 Feature
- 🦣 Legacy

closes #65


dralley:
Huh.  Wonder why the test fails, nothing looks wrong...

drahnr (approved):

PR #95, opened 2023-03-06
Streamline pull request template

dralley:
The current PR template is very thorough but a bit overwhelming and overkill for most contributions, slimming it down a bit might be helpful.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,45 +1,23 @@
-<!--
+<!---
 Thank you for submitting a PR to the rust rpm implementation!
--->
 
-# PR Content Desc
-
-<!---
-DELETE all that do not apply:
--->
-
-- 🩹 Bug Fix
-- 🦚 Feature
-- 📙 Documentation
-- 🦣 Legacy
-- 🪣 Misc
-
-<!---
-Mention the linked issue here.
-This will automatically close the issue once the PR is merged
-and creates a cross reference.
--->
-
-Closes #
-
-## Changes proposed by this PR
-
-<!---
-Tell the reviewer WHAT changed, and WHY it had to change
+Commits should be distinct and have a clear purpose, with messages
+which explain to the reviewer WHAT changed, and WHY it had to change
 and how the WHAT and the WHY tie together.
--->
 
-## Notes to reviewer
+At the bottom of your commit messages, add a line "Closes #IssueNumber)"
+for any issues resolved by the commit, substituting in an actual issue number.
+This will automatically close the issue once the PR is merged and creates a
+cross reference.
 
-<!---
-Leave a message to whoever is going to review this PR.
-Mainly, pointers to review the PR, and how they can test it.
 If things are still WIP or feedback on particular impl details
-are wanted, state them here too.
+are wanted, state them here or leave comments below.
 -->
 
-## 📜 Checklist
+### 📜 Checklist
 
+- [ ] Commits are cleanly separated and have useful messages
+- [ ] A changelog entry or entries has been added to CHANGELOG.md
+- [ ] Documentation is thorough
 - [ ] Test coverage is excellent and passes
 - [ ] Works when tests are run `--all-features` enabled
-- [ ] Documentation is thorough


.github/pull_request_template.md
```diff
@@ -1,45 +1,19 @@
-<!--
-Thank you for submitting a PR to the rust rpm implementation!
--->
-
-# PR Content Desc
```
dralley, line 5: If the user creates a PR with only a single commit, their commit message will be put at the top of the file, and then it needs to be copied down below.

.github/pull_request_template.md
```diff
@@ -1,45 +1,19 @@
...
-
-<!---
-DELETE all that do not apply:
--->
-
-- 🩹 Bug Fix
-- 🦚 Feature
-- 📙 Documentation
-- 🦣 Legacy
-- 🪣 Misc
```
dralley, line 15: Tags would be better for this I think

.github/pull_request_template.md
```diff
@@ -1,45 +1,19 @@
...
-
-<!---
-Mention the linked issue here.
-This will automatically close the issue once the PR is merged
-and creates a cross reference.
--->
-
-Closes #
```
dralley, line 23: Ideally a user puts this in their commit message in which case it will show up here automatically, and this line will need to be deleted.

.github/pull_request_template.md
drahnr: Tags cannot be added by external contributors

.github/pull_request_template.md
drahnr: Which unfortunately is rarely the case

.github/pull_request_template.md
drahnr: Commit messages tend to be more brief than a content description, I am generally trying to optimize for very little version inquiry from the maintainer pov while still managing a decent experience for non-professionals/junior devs that need more guidance.

drahnr (changes requested):
I disagree, the current PR template is text entirely and the intention is to give clear guidance on what a good PRentails for those that are not regulars, which is my personal crusade to make projects more accessible.

Most PRs from outside contributors are rarely atomic, so will be squashed anyways and mandate referencing the issue in the PR desc. Verbalizing the PR content helps reflect the author on the content and gives a hint of the intent. The emojis should be removed once labels can be put on by users without requiring the user going through a form style issue submission.

.github/pull_request_template.md
dralley: To be clear, I'm not suggesting that they omit the description, the comments below still request that.  Just that the description header (this specific line) unnecessary and gets in the way a little bit.

.github/pull_request_template.md
dralley: But how is asking them to do it there different from asking them to do it here?

.github/pull_request_template.md
dralley: What is the value add of these groups though?  If the goal is to make the contributor think about the changes they are making, then changelog entries already largely accomplish this in a more effective (IMO) way.

dralley:
>I disagree, the current PR template is text entirely and the intention is to give clear guidance on what a good PRentails for those that are not regulars, which is my personal crusade to make projects more accessible.

I totally agree with this goal, I'm just not sure it's as accessible in practice as the intention.  IMO these tweaks would improve the accessibility by e.g.

* Reducing the amount of prompt increases the probability that a contributor will actually read it regularly
* Reducing the need to rearrange sections of text (dropping "closes #" if there is no issue, fixing up the auto-added commit description if there is only one commit) removes annoyances
* Reducing the duplication of information prevents it from getting outdated and reduces the hoops to jump through

dralley:
@drahnr Left responses

.github/pull_request_template.md
drahnr: I see your point, but they all apply to changelogs as well.

.github/pull_request_template.md
drahnr: Because git doesn't ask them to :) this is an explicit request as part of the work flow to open a PR

.github/pull_request_template.md
```diff
@@ -1,45 +1,23 @@
...
-
-## Changes proposed by this PR
+Thank you for submitting a PR to the rust rpm implementation!
 
-<!---
-Tell the reviewer WHAT changed, and WHY it had to change
+Ideally, commits should be cleanly separated by purpose, with messages
```
drahnr, line 4: We should drop the `Ideally, `

drahnr (approved):
I see your points, the only thing I don't necessarily agree with, is the changelogs.

Imho changelogs should optimally be generated from the git history given one has sufficient description and descriptive title in each of them.

.github/pull_request_template.md
dralley: Well, the value in that case is mostly for the end consumers not the maintainers, but it's there for someone  :)

drahnr (approved):

PR #96, opened 2023-03-06
Added RPMPackage::open()

dralley:
Adds a helper method that avoids making the user do buffer management for common use cases.

- 🦚 Feature

### DIFF
--- a/src/rpm/package.rs
+++ b/src/rpm/package.rs
@@ -1,4 +1,5 @@
-use std::path::PathBuf;
+use std::io::BufReader;
+use std::path::{Path, PathBuf};
 
 use chrono::offset::TimeZone;
 #[cfg(feature = "async-futures")]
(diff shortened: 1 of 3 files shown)


src/rpm/package.rs
```diff
@@ -34,6 +35,12 @@ pub struct RPMPackage {
 }
 
 impl RPMPackage {
+    pub fn open<P: AsRef<Path>>(path: P) -> Result<Self, RPMError> {
+        let rpm_file = std::fs::File::open(path.as_ref())?;
```
drahnr, line 39: Could we add `fs_err` to avoid the lacking io error Info? I'd be much in favor of that.

drahnr: Doc comment would be good 😊

dralley: Are you referring to https://docs.rs/fs-err/latest/fs_err/?

drahnr (approved):

src/rpm/package.rs
drahnr: I think everything else should be `Reaser` and `Writer` trait bounded

dralley: There's a variant of RPMError for it, I'm not sure what is using it.

PR #99, opened 2023-03-07
Add parsing Compressor and add compress feature

ikrivosheev:
# PR Content Desc

- [x] Add support for parsing `Compressor` string in `RPMPackageMetadata`
- [x] Add Cargo feature for enable or disable compress feature 

- 🦚 Feature

Closes #

## 📜 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   and provenance of packages.
 - Added `get_package_segment_boundaries()` to `RPMPackage` to enable reading the raw bytes of the
   different components (header, payload, etc.) from an on-disk package.
+- Added `CompressionType`.
 
 ### Fixed
 
@@ -50,6 +51,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Removed `$pkg.metadata.get_payload_format()`. It is still possible to fetch manually, but
   practically speaking it is not meaningful. rpmbuild has written a misleading value here for
   10 years.
+- Added support for parsing `CompressionType` string in `RPMPackageMetadata`.
+- Changed signature for `RPMBuilder::compression`.
 
 ## 0.9.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `get_package_segment_boundaries()` to `RPMPackage` to enable reading the raw bytes of the
   different components (header, payload, etc.) from an on-disk package.
 - Added `CompressionType`.
+- Added support for `xz` compression type
 
 ### Fixed
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,7 +51,8 @@ chrono = "0.4"
 log = "0.4"
 itertools = "0.10"
 hex = { version = "0.4", features = ["std"] }
-zstd = "0.12.0"
+zstd = "0.12"
+xz2 = "0.1"
 futures = { version = "0.3.25", optional = true }
 
 # Libraries required for with_file_async() implementations
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -22,7 +22,7 @@
 //! # {
 //! let raw_secret_key = std::fs::read("./test_assets/secret_key.asc")?;
 //! let pkg = rpm::RPMBuilder::new("test", "1.0.0", "MIT", "x86_64", "some awesome package")
-//!             .compression(rpm::Compressor::from_str("gzip")?)
+//!             .compression(rpm::CompressionType::Gzip)
 //!             .with_file(
 //!                 "./test_assets/awesome.toml",
 //!                 rpm::RPMFileOptions::new("/etc/awesome/config.toml")
--- a/src/rpm/builder.rs
+++ b/src/rpm/builder.rs
@@ -18,9 +18,9 @@ use crate::sequential_cursor::SeqCursor;
 #[cfg(feature = "signature-meta")]
 use crate::signature;
 
-use crate::Digests;
 use crate::RPMPackage;
 use crate::RPMPackageMetadata;
+use crate::{CompressionType, Digests};
 
 #[cfg(feature = "with-file-async-tokio")]
 use tokio_util::compat::TokioAsyncReadCompatExt;
@@ -116,7 +116,7 @@ pub struct RPMBuilder {
     changelog_names: Vec<String>,
     changelog_entries: Vec<String>,
     changelog_times: Vec<chrono::DateTime<chrono::Utc>>,
-    compressor: Compressor,
+    compressor: CompressionType,
 
     vendor: Option<String>,
     url: Option<String>,
@@ -204,7 +204,7 @@ impl RPMBuilder {
         self
     }
 
-    pub fn compression(mut self, comp: Compressor) -> Self {
+    pub fn compression(mut self, comp: CompressionType) -> Self {
         self.compressor = comp;
         self
     }
@@ -583,13 +583,12 @@ impl RPMBuilder {
         // Lead is not important. just build it here
 
         let lead = Lead::new(&self.name);
-        let possible_compression_details = self.compressor.get_details();
 
-        // Make the borrow checker happy
-        let is_zstd = matches!(self.compressor, Compressor::Zstd(_));
         // Calculate the sha256 of the archive as we write it into the compressor, so that we don't
         // need to keep two copies in memory simultaneously.
-        let mut archive = Sha256Writer::new(&mut self.compressor);
+        let mut compressor: Compressor = self.compressor.try_into()?;
+        let possible_compression_details = compressor.get_details();
+        let mut archive = Sha256Writer::new(&mut compressor);
 
         let mut ino_index = 1;
 
@@ -674,7 +673,7 @@ impl RPMBuilder {
             "4.0-1".to_string(),
         ));
 
-        if is_zstd {
+        if matches!(self.compressor, CompressionType::Zstd) {
             self.requires.push(Dependency::rpmlib(
                 "rpmlib(PayloadIsZstd)".to_string(),
                 "5.4.18-1".to_string(),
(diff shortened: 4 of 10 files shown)


src/rpm/package.rs
```diff
@@ -324,9 +326,10 @@ impl RPMPackageMetadata {
     }
 
     #[inline]
-    pub fn get_payload_compressor(&self) -> Result<&str, RPMError> {
-        self.header
-            .get_entry_data_as_string(IndexTag::RPMTAG_PAYLOADCOMPRESSOR)
+    pub fn get_payload_compressor(&self) -> Result<Compressor, RPMError> {
+        let comp_str = self.header
+            .get_entry_data_as_string(IndexTag::RPMTAG_PAYLOADCOMPRESSOR)?;
+        Compressor::from_str(comp_str)
```
dralley, line 332: I'm not sure about this as it would be an exception from how it works for all other accessors, we're just returning the values in a mostly-raw form.  I understand the benefit of parsing to an enum type in many cases but I'm not so sure it applies strongly here?

Open to discussion on this point

Cargo.toml
```diff
@@ -60,7 +60,9 @@ tokio = {version = "1", features = ["full"]}
 tokio-util = { version = "0.7.4", features = ["compat"]}
 
 [features]
-default = ["signature-pgp", "async-futures", "with-file-async-tokio"]
+default = ["signature-pgp", "async-futures", "with-file-async-tokio", "compress"]
+gzip = ["libflate"]
+compress = ["gzip", "zstd"]
```
dralley, line 65: I do think it makes sense to split out these features, but I'm curious about your motivation, is it primarily just about code size / compilation time?  Or system dependencies that complicate cross-compilation?

I'm not sure about the "compress" feature as AFAIK nearly all RPMs use some form of compression, disabling it entirely is not very helpful.

src/rpm/package.rs
ikrivosheev: @dralley hello! Thank you for the review! I have some thoughts:

1. `mostly-raw` is `bytes` for me, but here is `&str`. We did some simple parsing before.
2. `crate` doesn't have API for working with this strings. I can only print information about Package but I cannot work with it.
3. I think `enum Compress` can have `Compress::Unknow(String)`  were we can store unknowing compress methods and didn't raise error

Cargo.toml
ikrivosheev: Well, I am writing `antivirus` and I need `rpm` package parser. In my project I use other library for ungziping.

drahnr: It sounds pretty fair to have 2 staged features, with all compression algorithms and individuals selectable

src/rpm/package.rs
drahnr: I prefer the enum, if only a limited set of strings are used/supported and raise an error otherwise.

ikrivosheev:
> @ikrivosheev do you want to update the PR, I'd be happy to integrate this rather sooner than later :)

Yes, I do. I will to it now)

ikrivosheev:
@drahnr done)

src/rpm/package.rs
ikrivosheev: Removed

Cargo.toml
ikrivosheev: Removed

drahnr:
It seems like the actual compression is missing now?

drahnr:
I misunderstood, all good

drahnr (approved):

src/tests/mod.rs
```diff
@@ -89,7 +89,10 @@ fn test_rpm_header_base(package: RPMPackage) -> Result<(), Box<dyn std::error::E
     );
     assert_eq!(package.metadata.get_build_time().unwrap(), 1540945151);
 
-    assert_eq!(package.metadata.get_payload_compressor().unwrap(), "xz");
+    assert!(matches!(
+        package.metadata.get_payload_compressor(),
+        Err(RPMError::UnknownCompressorType(_))
```
dralley, line 94: Here and below you should match `Err(RPMError::UnknownCompressorType("xz".to_string()))` instead

dralley:
Should get a changelog entry.  Also, please squash the commits and update the commit message to reflect what it is currently doing, which isn't what it was originally doing

src/tests/mod.rs
ikrivosheev: Done

ikrivosheev:
> Should get a changelog entry. Also, please squash the commits and update the commit message to reflect what it is currently doing, which isn't what it was originally doing

Done

src/rpm/package.rs
dralley: So, as currently implemented it looks like this returns not a compression type, but a `Compressor` instance, which feels like it would be kind of awkward to match against, because as a user I'm never going to use it as a compressor and I don't care about the associated data, just the tags.

drahnr: I'd tend to say yes, but I think anything typed is better than strings.

dralley: >It's kinda nuts that we expose the current Compressor type publicly for the builder pattern.

Well not entirely as you might want to e.g. tweak the compression level, and we provide no other way to do that.  But long term I agree it's not ideal.

ikrivosheev: Well, ok. I agree with you. I will push changes today or tomorrow)

dralley: @ikrivosheev Reminder :)

ikrivosheev: @dralley I remember! Sorry for the long answer!

ikrivosheev: Done

CHANGELOG.md
```diff
@@ -43,6 +43,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Removed `$pkg.metadata.get_payload_format()`. It is still possible to fetch manually, but
   practically speaking it is not meaningful. rpmbuild has written a misleading value here for
   10 years.
+- Added `CompressionType`
```
dralley, line 46: This isn't a breaking change, move it up to the `Added` section

ikrivosheev: Ups, thank you! Done.

ikrivosheev:
@dralley @drahnr  I have done. Can you review the PR. I hope I got the idea right

src/rpm/builder.rs
```diff
@@ -591,13 +591,14 @@ impl RPMBuilder {
         // Lead is not important. just build it here
 
         let lead = Lead::new(&self.name);
-        let possible_compression_details = self.compressor.get_details();
 
         // Make the borrow checker happy
-        let is_zstd = matches!(self.compressor, Compressor::Zstd(_));
+        let is_zstd = matches!(self.compressor, CompressionType::Zstd);
```
dralley, line 596: We don't need to make the borrow checker happy anymore ^.

ikrivosheev: Done

src/lib.rs
```diff
@@ -22,7 +22,7 @@
 //! # {
 //! let raw_secret_key = std::fs::read("./test_assets/secret_key.asc")?;
 //! let pkg = rpm::RPMBuilder::new("test", "1.0.0", "MIT", "x86_64", "some awesome package")
-//!             .compression(rpm::Compressor::from_str("gzip")?)
+//!             .compression(rpm::CompressionType::Gzip)
```
dralley, line 25: `CompressionDetails` currently uses `&str` for `compression_type` and `compression_level`, it should use `CompressionType` for the former and `u8` for the latter.

What we're missing here is some way for the user to specify the compression level to use.  `CompressionDetails` would actually be nice for this.

So how about we implement the signature like this: `fn compressor<T: Into<CompressionDetails>>(details: T)` and implement `Into<CompressionDetails> for CompressionType` with the default values.  And then a function implemented on `CompressionType` such as:

```
CompressionType::with_type(level: u8)
```

with some checks to make sure it's a reasonable value

`Compressor` doesn't need to be public anymore, just `pub(crate)`

src/rpm/compressor.rs
```diff
@@ -1,25 +1,47 @@
 use crate::errors::*;
 use std::io::Write;
 
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Default)]
+pub enum CompressionType {
+    #[default]
+    None,
+    Gzip,
+    Zstd,
```
dralley, line 9: Go ahead and add `Xz` here

src/rpm/package.rs
```diff
@@ -591,9 +591,11 @@ impl RPMPackageMetadata {
     }
 
     #[inline]
-    pub fn get_payload_compressor(&self) -> Result<&str, RPMError> {
-        self.header
-            .get_entry_data_as_string(IndexTag::RPMTAG_PAYLOADCOMPRESSOR)
+    pub fn get_payload_compressor(&self) -> Result<CompressionType, RPMError> {
+        let comp_str = self
```
dralley, line 595: If this is the only place where `CompressionType::from_str()` is used then lets inline it here.

In the case of an uncompressed package, the tag simply won't exist.  So we could either return the error, or match on the result and return `CompressionType::None` in that case.  @drahnr What do you prefer?

src/lib.rs
ikrivosheev: @dralley, it's a good suggestion! Well, I will try to create small POC and will post it here.

src/rpm/package.rs
ikrivosheev: @dralley I suggest something like this:

```
enum CompressionType {
    None, 
    ...,
    Unknown(String)
```

What do you think about this variant?

drahnr: @ikrivosheev I don't have a strong opinion, I am against inlining `FromStr` though.

src/rpm/compressor.rs
ikrivosheev: @dralley hello. I added support `xz`.

src/lib.rs
ikrivosheev: @dralley well, I have one problem. I cannot create function with signature: `fn compressor<T: Into<CompressionDetails>>(details: T)`. Only: `fn compressor<T: TryInto<CompressionDetails>>(details: T) -> Result<(), RPMError>`. If this is ok I will push changes.

ikrivosheev: @dralley, hello! Can you look your idea in my implementation: https://github.com/rpm-rs/rpm/pull/135

drahnr (approved):

dralley (approved):
I can live with this, I might want to make adjustments later but they wouldn't change the API

PR #100, opened 2023-03-08
Stop store content in memory for RPMBuilder

ikrivosheev:
# PR Content Desc

At the moment, `RPMPackage` store content in memory. This is not good situation. I rewrite `RPMPackage` to generic struct.

If my idea is ok for maintainers, I will finish the PR.

- 🦚 Feature
- 📙 Documentation
- 🦣 Legacy
- 🪣 Misc

Closes #

## 📜 Checklist

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -22,7 +22,7 @@
 //! let raw_secret_key = std::fs::read("./test_assets/secret_key.asc")?;
 //! // It's recommended to use timestamp of last commit in your VCS
 //! let source_date = 1_600_000_000;
-//! let pkg = rpm::RPMBuilder::new("test", "1.0.0", "MIT", "x86_64", "some awesome package")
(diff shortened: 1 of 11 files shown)


dralley:
I'm going to try to take a step in the right direction with a PR to address https://github.com/rpm-rs/rpm/issues/107

A general complication is that

* you have to generate the payload in order to calculate a checksum of the payload
* you have the calculate a checksum of the payload in order to put that checksum in the header (also other attributes like the size of the payload)
* you have to have a completed header before you can assemble and sign the package

So the first step is just going to be to avoid storing everything twice, once independently in the list of file entries to be processed, and once in the contents of the payload.

ikrivosheev:
Well, good news) Some tests work now)

dralley:
I would hold off on continuing this for the time being.  I think there are some interesting ideas here, but I'm not sure about the implementation.  There are a couple of bigger problems to address first when it comes to builder memory use, and on the reading side, the introduction of `RPMPackageMetadata` mostly handles with that problem.

PR #101, opened 2023-03-09
Remove "Lead" from the public API and stop checking its fields

### LINKED ISSUE #78: Don't rely on the existence of the "Lead", and don't populate it
See https://github.com/rpm-software-management/rpm/issues/2368

dralley:
# PR Content Desc

Lead has been deprecated upstream for a long time and the docs suggest not to read it.  There is a plan to stop populating its fields entirely, so we should stop trying to validate them.

closes #78

- 🦣 Legacy

## Notes to reviewer

See the references attached to the issue

## 📜 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Bump MSRV to 1.63.0
 - Removed async support from default features
+- Removed `Lead` from the public API. `Lead` is long-deprecated and shouldn't be relied on.
+  Restricted the data we write to `Lead` to the bare minimum required for compatibility.
 
 ## 0.9.0
 
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -49,21 +49,6 @@ pub enum RPMError {
     #[error("invalid tag value enum variant for {tag} with {variant}")]
     InvalidTagValueEnumVariant { tag: String, variant: u32 },
 
-    #[error("unsupported lead major version {0} - only version 3 is supported")]
-    InvalidLeadMajorVersion(u8),
-
-    #[error("unsupported lead major version {0} - only version 0 is supported")]
-    InvalidLeadMinorVersion(u8),
-
-    #[error("invalid type - expected 0 or 1 but got {0}")]
-    InvalidLeadPKGType(u16),
-
-    #[error("invalid os-type - expected 1 but got {0}")]
-    InvalidLeadOSType(u16),
-
-    #[error("invalid signature-type - expected 5 but got {0}")]
-    InvalidLeadSignatureType(u16),
-
     #[error("invalid size of reserved area - expected length of {expected} but got {actual}")]
     InvalidReservedSpaceSize { expected: u16, actual: usize },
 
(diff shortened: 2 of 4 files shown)


dralley:
Current versions of `rpm` still look at these fields apparently though they perhaps shouldn't, which is why the tests fail.  So we may have to hold off until there is a current version of RPM which is forwards compatible with that aspect of RPMv6.

https://github.com/rpm-software-management/rpm/issues/2368#issuecomment-1461320583

dralley:
Actually what this probably means is that we just can't merge this, because in practice the RPMs will simply not be installable on important platforms.  At least, not without some kind of legacy fallback mode.

src/rpm/builder.rs
```diff
@@ -67,10 +67,27 @@ async fn async_file_mode(_file: &async_std::fs::File) -> Result<u32, RPMError> {
...
 }
 
+/// Compatibility profile
+///
+/// Older versions of RPM require certain feature flags to be enabled whereas
+/// newer versions assume their presence. We default to the `Legacy` profile for
+/// the widest possible compatibility.
+///
+/// TODO: this is internal-only (basically dead code) until the plan for "RPM v6"
+/// has been finalized upstream.
```
dralley, line 77: @drahnr Options:

1) We can defer all of this until later, and I remove this code.  This PR gets pared down to just the removal of "Lead" from the API.
2) We can leave it like this, even though it's dead code, so that the some of the low-stakes changes can be worked on ahead of time
3) We can do #2, but via some different mechanism (like compile-time feature selection) instead of runtime configuration

drahnr: Depending how far the file formats align, different parsers while re-using data structures might be the best choice. Then hiding the different implementations behind a feature flag is secondary to that.

drahnr: I only skimmed the rpm V6 Format, but it doesn't appear to be too radically different. So the above outline approach should work.

drahnr: Since might be a while until rpm V6 hits the ground, and we don't want to be in the mess of not supporting existing installations when updating a tool that builds on-top of rpm-rs/rpm, we should be wary of the side effects and fallout. Even if it would be major server bump, we should always support more rather than less (or at least with a lot of lead time).

dralley: Given the on-disk format isn't changing I don't think we need separate parsers.  Actually parsers in general don't seem problematic, it's more the writing of the RPMs that is.  We'd like to be able to produce "clean" modern RPMs without the legacy baggage, but we still need to support the baggage for users building e.g. CentOS 7 packages.

dralley:
For now, I've restricted this to just the changes in the title

drahnr (approved):

drahnr:
What are the current magic vales? Are those the ones used? And are we sure, we can even parse older packages? If so, LGTM, otherwise I'd refrain from merging

dralley:
> What are the current magic vales? Are those the ones used? 

Well there's "magic values" and then there's the "magic number".  Magic number is just the first few bytes of the file and is just used for file type recognition.  Otherwise, the values currently written are the same ones RPM uses currently.

PR #102, opened 2023-03-12
Fix rpmlib() requires

dralley:
# PR Content Desc

Add the appropriate `rpmlib()` requires to built packages.  Fixed up the README example code

- 🩹 Bug Fix

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 [![crates.io](https://img.shields.io/crates/v/rpm.svg)](https://crates.io/crates/rpm)
-[![docs.rs](https://docs.rs/rom/badge.svg)](https://docs.rs/rpm)
+[![docs.rs](https://docs.rs/rpm/badge.svg)](https://docs.rs/rpm)
 [![MSRV](https://img.shields.io/badge/rustc-1.63.0+-ab6000.svg)](https://blog.rust-lang.org/2022/08/11/Rust-1.63.0.html)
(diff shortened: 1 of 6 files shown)


src/rpm/builder.rs
```diff
@@ -1023,34 +1045,6 @@ impl RPMBuilder {
...
-        //     "4.0-1".to_string(),
-        // ));
-
-        // self.requires.push(Dependency::rpm_lib(
-        //     "rpmlib(CompressedFileNames)".to_string(),
-        //     "3.0.4-1".to_string(),
-        // ));
-
-        // self.requires.push(Dependency::rpm_lib(
-        //     "rpmlib(PayloadIsXz)".to_string(),
```
dralley, line 1046: This might be the part that was breaking.  Obviously the payload is not XZ because we don't support it.

README.md
```diff
@@ -32,6 +32,8 @@ This library does not build software like rpmbuild. It is meant for finished art
 ### Examples
 
 ```rust
+use std::str::FromStr;
```
drahnr, line 35: Isn't this part of the prelude?

drahnr (approved):

README.md
dralley: I don't think so.  In any case I was getting some compile errors without it when I added an example.

I removed the example because we already have one in `lib.rs` and having the same code in 3 places is excessive.  But it's there too.

PR #106, opened 2023-03-16
Always create package digests

### LINKED ISSUE #105: Package checksums are not calculated during build when "signature-meta" feature is disabled
Not verifying digests during parsing is one thing, but them to not be created during package build feels like maybe a step too far?  It means there would be no internal mechanism for detecting corruption of the payload.

https://github.com/rpm-rs/rpm/blob/86b9bd13c31df0bedebd2ad190ec84d1256bf152/src/rpm/builder.rs

dralley:
Remove the signature-meta feature because it's a bit superfluous

closes #105

### 📜 Checklist


src/compat_tests.rs
```diff
@@ -18,14 +18,12 @@ fn cargo_out_dir() -> std::path::PathBuf {
...
 
-#[cfg(feature = "signature-meta")]
-use signature::{self, Verifying};
-
 #[cfg(feature = "signature-pgp")]
 mod pgp {
     use super::*;
     use futures::io::AsyncWriteExt;
     use signature::pgp::{Signer, Verifier};
+    use signature::{self, Verifying};
```
drahnr, line 26: Should always be exported if we remove `"signature-meta"` - after all it was meant for the case where signature creation/verification is desired, but not the internal implementations.

drahnr (changes requested):
See the inline comment. I think we should decouple signature creation from the signing. Hence I don't think combining `pgp` with the API of signing/verification is the right approach.

src/compat_tests.rs
dralley: This is in a tests file, it'll just be an unused import.

src/rpm/headers/header.rs
```diff
@@ -308,7 +308,7 @@ impl Header<IndexSignatureTag> {
     /// PGP and RSA tags expect signatures according to [RFC2440](https://tools.ietf.org/html/rfc2440)
     ///
     /// Please use the [`builder`](Self::builder()) which has modular and safe API.
-    #[cfg(feature = "signature-meta")]
+    #[allow(unused)]
     pub(crate) fn new_signature_header(
```
dralley, line 312: Isn't this decoupling the signing from the signature creation?  I'm not 100% sure I understand what you mean.

src/compat_tests.rs
dralley: If you're talking about the other module in general, I removed the feature flag and made it public.  The `pgp` submodule is still behind the feature flag.

dralley: Never mind, I think I see what you mean.

dralley:
@drahnr Take another look, I removed `[cfg(feature = "signature-pgp")]` from everything not explicitly touching `pgp`

drahnr (changes requested):
It's not in going into the direction I was thinking. Why change all the feature flags regarding signatures? I'd have expected splitting the hash/checksum creation from the whole signature creation logic, which does not need any changes on the feature flags side.

drahnr:
I'll comment in more detail at some point in the next few days

dralley:
@drahnr ^

src/rpm/builder.rs
```diff
@@ -389,7 +387,6 @@ impl RPMBuilder {
         let mut header = Vec::with_capacity(128);
         header_idx_tag.write(&mut header)?;
 
-        #[cfg(feature = "signature-meta")]
```
drahnr, line 392: Yes, we should always build this, so removing the signature-meta tag here is correct.

src/rpm/builder.rs
```diff
@@ -422,7 +417,6 @@ impl RPMBuilder {
     /// use an external signer to sing and build
     ///
     /// See `signature::Signing` for more details.
-    #[cfg(feature = "signature-meta")]
```
drahnr, line 425: This is where I disagree. `build_and_sign` is user visible API and should not be usable for skipping signatures. Imho `build` should also create the header, but not sign it.

src/rpm/builder.rs
```diff
@@ -471,7 +465,6 @@ impl RPMBuilder {
     }
 
     /// use prepared data but make sure the signatures are
-    #[cfg(feature = "signature-meta")]
```
drahnr, line 474: :heavy_check_mark:

src/rpm/headers/header.rs
```diff
@@ -163,7 +163,7 @@ where
             .ok_or_else(|| RPMError::TagNotFound(tag.to_string()))
     }
 
-    #[cfg(feature = "signature-meta")]
+    #[allow(unused)]
```
drahnr, line 166: It shouldn't be unused, either it's required in meta or any of the the given concrete signer implementations which should end up in a `cfg(any(..))` annotation or be deleted.

src/tests.rs
```diff
@@ -31,7 +31,7 @@ fn cargo_manifest_dir() -> std::path::PathBuf {
     std::path::PathBuf::from(env!("CARGO_MANIFEST_DIR"))
 }
 
-#[cfg(feature = "signature-meta")]
+#[cfg(feature = "signature-pgp")]
```
drahnr, line 34: Doesn't require the pgp signer, the API is enough.

drahnr (changes requested):

src/rpm/headers/header.rs
dralley: The reason it's unused is that the `get_entry_data_as_binary()` function is only used by the signature verification code, which is only enabled by `signature-pgp`, and because we don't currently have any public functions for accessing header data directly (they're all `pub(crate)`).  Probably we *should* have those public functions (as we can't predict every single tag a user may want to be able to read), and then it wouldn't be unused.

src/rpm/builder.rs
dralley: >This is where I disagree. `build_and_sign` is user visible API and should not be usable for skipping signatures.

What do you mean by "not be usable for skipping signatures"?  You have to pass an `S: signature::Signing<crate::signature::algorithm::RSA>` in order to use this API, so it can't be skipped.  For most users, that means it won't be very useful without enabling `signature-pgp`, but since we've discussed users writing their own implementation of this trait on other issues, that's not universally true.  

>Imho `build` should also create the header, but not sign it.

That's what it currently does (in this PR)?  It produces a signature header and populates it with digests, but not the signatures.

dralley: On a separate note though, we could almost do away with `build_and_sign()` if we take `RPMPackage::sign()` and make it return `self` instead of `()`.

That way you could write this chain, which ought to be equivalent to `build_and_sign()`.

```
let pkg = RpmBuilder::new(....)
    [....]
    .build()
    .sign(signer);
```

drahnr:
As you can imagine from my latency, I didn't get around to suggest a viable path, but I do have an intuition what needs to be done. I hope I can create a draft PR really soon

drahnr:
Closing, #111 is merged and covers the same functionality, but does not interleave it with the signature logic but detaches it.

PR #109, opened 2023-03-23
Drop cpio-rs, replace with vendored code

### LINKED ISSUE #108: Drop "cpio" libraries and write something semi-custom, because RPM doesn't use vanilla CPIO
See the "Payload" section of the website: https://rpm-software-management.github.io/rpm/manual/format.html

> # Payload
> 
> The Payload is currently a cpio archive, gzipped by default. The cpio archive type used is SVR4 with a CRC checksum.
> 
> As cpio is limited to 4 GB (32 bit unsigned) file sizes RPM since version 4.12 uses a stripped down version of cpio for packages with files > 4 GB. This format uses 07070X as magic bytes and the file header otherwise only contains the index number of the file in the RPM header as 8 byte hex string. The file metadata that is normally found in a cpio file header - including the file name - is completely omitted as it is stored in the RPM header already.

So, we should fork [`cpio-rs`](https://github.com/jcreekmore/cpio-rs) (providing the appropriate credits of course), strip it down to the subset we need, and change the magic bytes constant.

Luckily the CPIO format is pretty simple and the library only a few hundred lines, so it's not a big deal.

Subsequently we need to change the `PAYLOADFORMAT` tag, but upstream RPM still uses `cpio` as the name, so we'll have to wait until they pick something.

dralley:
This is necessary because RPM uses a modified CPIO format to support
\>4gb file sizes. To maximize similarity and avoid producing nonsensical
RPM payloads, we should do the same.

closes #108

### 📜 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added support for ecdsa signatures
+- Added `Package::files()` for iterating over the files of an RPM package (metadata & contents).
 
 ## 0.16.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `FileOptions::verify()`
 - Added `Evr` and `Nevra` structs and `rpm_evr_compare` function for comparing RPM versions.
 
+### Fixed
+
+- RPM packages that use large files (>4gb) now correctly declare rpmlib() dependency
+
 ### Changed
 
 - As RHEL 7 (thus, CentOS 7 and other derivatives) goes out-of-support on June 30, 2024, support for legacy
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,7 +53,7 @@ itertools = "0.13"
 hex = { version = "0.4", features = ["std"] }
 zstd = { version = "0.13", optional = true }
 xz2 = { version = "0.1", optional = true }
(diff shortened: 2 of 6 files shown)


src/rpm/payload/README.txt
```diff
@@ -0,0 +1,11 @@
+This code is copied from the crate `cpio-rs` by Jonathan Creekmore (https://github.com/jcreekmore/cpio-rs).
```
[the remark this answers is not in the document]
cmeister2: Think you're ok to have the LICENSE.txt next to it.

src/rpm/payload/mod.rs
```diff
@@ -0,0 +1,550 @@
+//! Read/write `newc` (SVR4) format archives.
```
cmeister2, line 1: I don't mind you vendoring the code, but it's impossible to see the changes you've made on top of the original code. For suggestion only: split the code into two commits; one with the original code, one with your changes on top.

dralley: Way ahead of you, this is that initial commit, no changes have actually been added yet

cmeister2: Perfect 😀

dralley: Yeah the errors are because of all the dead code

drahnr:
Could you outline the exact differences/changes required to make the existing CPIO work?

dralley:
@drahnr I outline it in https://github.com/rpm-rs/rpm/issues/108

dralley:
Note to self: make sure we handle the edge cases discussed here: https://blog.colindou.ch/posts/lets-make-an-os-cpio-weirdness/

That is: be careful with directories, and be careful with absolute paths.

PR #110, opened 2023-04-01
Add new helper functions to package metadata & header

dralley:
### 📜 Checklist


src/rpm/headers/header.rs
```diff
@@ -249,7 +249,7 @@ where
             })
     }
 
-    pub(crate) fn get_entry_data_as_string_array(&self, tag: T) -> Result<&[String], RPMError> {
+    pub fn get_entry_data_as_string_array(&self, tag: T) -> Result<&[String], RPMError> {
```
drahnr, line 252: Do we really want these low level methods to be exposed?

src/rpm/package.rs
```diff
@@ -313,14 +314,137 @@ impl RPMPackageMetadata {
...
+    fn get_dependencies(
+        &self,
+        names_tag: IndexTag,
+        flags_tag: IndexTag,
+        versions_tag: IndexTag,
+    ) -> Vec<Dependency> {
+        // Return an empty list if the tags are not present
+        if self.header.find_entry_or_err(&names_tag).is_err() {
+            return vec![];
+        }
```
drahnr, line 326:
```suggestion
        let names = if let Ok(names) = self.header.get_entry_data_as_string_array(&names_tag) {
            names
        } else {
        return vec![];
        };
```

src/rpm/package.rs
```diff
@@ -313,14 +314,137 @@ impl RPMPackageMetadata {
...
 
-    #[inline]
-    pub fn get_payload_format(&self) -> Result<&str, RPMError> {
-        self.header
-            .get_entry_data_as_string(IndexTag::RPMTAG_PAYLOADFORMAT)
+        let names = self
+            .header
+            .get_entry_data_as_string_array(names_tag)
+            .unwrap();
```
drahnr, line 331:
```suggestion
```

src/rpm/package.rs
```diff
@@ -313,14 +314,137 @@ impl RPMPackageMetadata {
...
+        let flags = self.header.get_entry_data_as_u32_array(flags_tag).unwrap();
+        let versions = self
+            .header
+            .get_entry_data_as_string_array(versions_tag)
+            .unwrap();
```
drahnr, line 336:
```suggestion
            .ok_or_else(|| RpmError::...)?;
```

src/rpm/package.rs
```diff
@@ -313,14 +314,137 @@ impl RPMPackageMetadata {
...
+                |mut acc, (name, flags, version)| {
+                    acc.push(Dependency {
+                        dep_name: name.to_owned(),
+                        sense: flags,
+                        version: version.to_owned(),
+                    });
+                    Ok(acc)
+                },
+            )
+            .unwrap()
```
drahnr, line 350:
```suggestion
        Vec::from_iter(itertools::multizip((names, flags, versions))
            .map::(
                |(name, flags, version)| {
                    Dependency {
                        dep_name: name.to_owned(),
                        sense: flags,
                        version: version.to_owned(),
                    })
            )
```

src/rpm/package.rs
```diff
@@ -313,14 +314,137 @@ impl RPMPackageMetadata {
...
+    ///
+    /// "rpm" itself will ignore such dependencies, but a dependency solver may elect to treat this
+    /// package as a "requires" when the matching package is installed. Unlike a "requires" however,
+    /// if installing it would cause errors, it can be ignored ignored without error.
+    pub fn get_supplements(&self) -> Vec<Dependency> {
+        self.get_dependencies(
+            IndexTag::RPMTAG_SUPPLEMENTNAME,
+            IndexTag::RPMTAG_SUPPLEMENTFLAGS,
+            IndexTag::RPMTAG_SUPPLEMENTVERSION,
+        )
```
drahnr, line 447: :+1:

src/tests.rs
```diff
@@ -259,7 +402,6 @@ fn test_rpm_header_base(package: RPMPackage) -> Result<(), Box<dyn std::error::E
         assert_eq!(*tag, metadata.signature.index_entries[i].tag);
     }
 
-    assert_eq!("cpio", metadata.get_payload_format()?);
```
drahnr, line 262: Tbh, never encountered a single rpm file in the last decade where the content was not a CPIO archive :sweat_smile:

drahnr (changes requested):
A few code simplifications pending, mostly around unwrap and using patterns to avoid such. Also raising the question if we want to expose the low level~ish APIs exposing raw~ish data?

src/rpm/headers/header.rs
dralley: I think we probably do.  I mean there's just no way that we can expose helper functions for every single possible tag that someone somewhere might care about.

src/tests.rs
dralley: Technically all of them produced in the past decade (with recent versions of RPM) are not CPIO archives, the payload format tag is just lying.

https://github.com/rpm-software-management/rpm/issues/2447

src/tests.rs
```diff
@@ -401,6 +543,16 @@ async fn test_rpm_builder_async() -> Result<(), Box<dyn std::error::Error>> {
     Ok(())
 }
 
+// #[test]
+// fn test_rpm_builder_nofiles() -> Result<(), Box<dyn std::error::Error>> {
```
dralley, line 547: This test fails currently because the implementation expects these tags to always be present.

dralley:
@drahnr See the last commit named "temp".  Is that an acceptable direction?  I can also potentially generalize that to a macro.

src/rpm/package.rs
```diff
@@ -232,7 +232,7 @@ impl RPMPackageMetadata {
     #[inline]
     pub fn is_source_package(&self) -> bool {
         self.header
-            .get_entry_data_as_u32(IndexTag::RPMTAG_SOURCEPACKAGE)
+            .find_entry_or_err(&IndexTag::RPMTAG_SOURCEPACKAGE)
```
drahnr, line 235: Change in behavior, the presence of the key is now sufficient, previously it was it's value. Is that intended?

src/tests.rs
drahnr: Oh ffs..

src/tests.rs
drahnr: Should we expect that? I think we should support marker packages too, so I'd prefer if this is not disabled, but fixed.

src/rpm/package.rs
dralley: Yes

src/tests.rs
dralley: I'm just pointing out that we ought to do for the other iterators what we have done for the dependency-getters, and return empty vecs, but I'd like feedback on the approach before copying it over.  Agree that it's a bug that should be fixed

dralley:
@drahnr Do you prefer the approach in temp or the approach in temp2?

src/rpm/package.rs
```diff
@@ -314,137 +314,124 @@ impl RPMPackageMetadata {
...
-        versions_tag: IndexTag,
-    ) -> Vec<Dependency> {
-        // Return an empty list if the tags are not present
-        if self.header.find_entry_or_err(&names_tag).is_err() {
-            return vec![];
-        }
+    fn get_dependencies(&self, tags: &'static [IndexTag; 3]) -> Result<Vec<Dependency>, RPMError> {
+        let names = self.header.get_entry_data_as_string_array(tags[0]);
+        let flags = self.header.get_entry_data_as_u32_array(tags[1]);
+        let versions = self.header.get_entry_data_as_string_array(tags[2]);
```
drahnr, line 320: Why not just bail early here and use `map_err(Error::TagCoexistenceMismatch)` - if we need the error map at all.

drahnr:
> @drahnr Do you prefer the approach in temp or the approach in temp2?

Neither, I'd prefer to handle the errors as they appear rather than moving `Result` types around to then check them all at once and create hard to read code.

dralley:
>     The first one (temp), but I'd prefer to handle the errors as they appear rather than moving Result types around to then check them all at once and create hard to read code.

You misunderstand.  If none of the tags exist, that is not an issue - packages are not required to have dependencies and if they do not then those tags will not exist.  If they do not exist, then you want to return an empty Vec, not an error.

If one of the tags exists but not the others then that is an error that ought to be raised.  You cannot know this before you've checked all of them, though.

drahnr:
> 

Yeah, I misread, in that case the second one with an explicit error variant.

dralley:
Unrelated question: is it actually useful to get a list of file checksums or signatures, without the paths they map to?  I question whether we actually need anything other than `get_file_entries()`.

dralley:
Dropping the explicit error variant for reasons laid out in https://github.com/rpm-rs/rpm/pull/110#issuecomment-1496078640

If we come up with a way around the limitations then we can restore it.

drahnr (approved):
I'd prefer to commit generated roms so checks don't depend on any local tools, such as `rpmbuild`.

Generally looks good, the scope extended a bit as the PR progressed

cmeister2:
Taking a look.

src/rpm/headers/header.rs
```diff
@@ -161,10 +161,11 @@ where
             .iter()
             .find(|entry| &entry.tag == tag)
             .ok_or_else(|| RPMError::TagNotFound(tag.to_string()))
+        // @todo: this could be more efficient, if the tag is an integer, we can just pass around
+        // an integer, and the name of the tag (or "unknown") can be easily derived from that
     }
 
-    #[cfg(feature = "signature-meta")]
-    pub(crate) fn get_entry_data_as_binary(&self, tag: T) -> Result<&[u8], RPMError> {
+    pub fn get_entry_data_as_binary(&self, tag: T) -> Result<&[u8], RPMError> {
```
cmeister2, line 168: Just wondering if there's a more efficient way to write this set of `get_entry_data` functions. Would it make more sense to have `get_entry_data` as a method, which returns an object that can convert into a binary array, a string, etc? Then the `EntryData` object could implement a bunch of standard traits (e.g. TryInto<&str>).

cmeister2: It would change the calling code to something like

```
self.header
            .get_entry_data(IndexTag::RPMTAG_DESCRIPTION)?
            .as_i18n_string()
```
or even `.try_into()`

src/rpm/package.rs
```diff
@@ -426,52 +582,113 @@ impl RPMPackageMetadata {
...
-                    path,
-                    ownership: FileOwnership {
-                        user: user.to_owned(),
-                        group: group.to_owned(),
+        match (modes, users, groups, digests, mtimes, sizes, flags) {
+            (Ok(modes), Ok(users), Ok(groups), Ok(digests), Ok(mtimes), Ok(sizes), Ok(flags)) => {
+                let paths = self.get_file_paths()?;
+                let n = paths.len();
+
+                let v = itertools::multizip((
```
cmeister2, line 597: Is it worth adding comments to this bit of code? Just to explain what it's trying to do.

src/tests.rs
```diff
@@ -125,9 +124,157 @@ fn test_rpm_header_base(package: RPMPackage) -> Result<(), Box<dyn std::error::E
     assert_eq!(package.metadata.get_build_time().unwrap(), 1540945151);
 
     assert_eq!(package.metadata.get_payload_compressor().unwrap(), "xz");
-    assert_eq!(package.metadata.get_payload_format().unwrap(), "cpio");
+    // @todo: too many to test for, need a new fixture RPM. it works though.
+    assert_eq!(
```
cmeister2, line 128: `assert!()` can be used for boolean checking.

cmeister2 (commented):
Looks generally good.

dralley:
>I'd prefer to commit generated roms so checks don't depend on any local tools, such as rpmbuild.

Yup, I did commit the generated RPMs, but I like having the ability to regenerate them at-will.  The way it is currently structured, you can regenerate them without automatically replacing them.

src/rpm/headers/header.rs
dralley: Well that's kind of what `find_entry_or_err()` is, yeah?

We'll probably have to tweak it a bit anyway because a proper implementation of `i18n` functions require reading header tags besides the one you're actively looking at. Hopefully before the next release to avoid API churn.

src/rpm/package.rs
```diff
@@ -344,37 +483,54 @@ impl RPMPackageMetadata {
     /// Extract a the set of contained file names.
     pub fn get_file_paths(&self) -> Result<Vec<PathBuf>, RPMError> {
         // reconstruct the messy de-constructed paths
-        let base = self
+        let basenames = self
             .header
-            .get_entry_data_as_string_array(IndexTag::RPMTAG_BASENAMES)?;
+            .get_entry_data_as_string_array(IndexTag::RPMTAG_BASENAMES);
         let biject = self
```
dralley, line 489: I have no idea what this name is supposed to be mean, do either of you know?

dralley:
Tweaked the script to disable compression, so that we can inspect the payload more easily, and so that the digests don't get broken by minor tweaks to the compression algorithms or the level used.

src/rpm/package.rs
dralley: I think it's fairly clear, but I wouldn't complain about more comments being added

dralley:
@drahnr Unfortunately I need a new approval because of the push.

Would you mind if I loosen the branch protection a tiny bit, so that tiny changes like the ones above can be made without requiring a fresh approval?

drahnr:
> @drahnr Unfortunately I need a new approval because of the push.
> 
> Would you mind if I loosen the branch protection a tiny bit, so that tiny changes like the ones above (or addressing comments) can be made without requiring a fresh approval?

Yes please, go ahead!

drahnr (approved):

PR #111, opened 2023-04-01
always digest

drahnr:
@dralley includes the part of your PR, some additional cleanups, and the direction of using a `fn verify_digest` explicitly, so the user can opt-in explicitly to have either or both checked. The `fn verify_digest` is _not_ feature gated, while the `fn verify_signature` is.

I hope this clarifies the conversation had in #106

Potentially I'll complete this PR soon™


dralley:
FYI, I'm working on an overhaul of the tests right now now, so it might be best to drop that commit - but I will try to introduce a similar split.

drahnr:
@drally please hold off, I'll complete these changes early next week

dralley:
Well, I was hoping to finish mine tomorrow or by the end of the weekend...

Note that I'm talking exclusively about the tests and not the digest PR.  It's separate.

dralley:
I don't, in general, like keeping tests that could (and ought to be) integration tests inside `src/`.  The tests commit goes further in that direction and so I'd prefer we didn't do that.  Just letting you know before you put too much work into it.

src/rpm/headers/header.rs
```diff
@@ -136,17 +136,17 @@ where
         Ok(())
     }
 
-    pub(crate) fn find_entry_or_err(&self, tag: &T) -> Result<&IndexEntry<T>, RPMError> {
+    pub(crate) fn find_entry_or_err(&self, tag: T) -> Result<&IndexEntry<T>, RPMError> {
```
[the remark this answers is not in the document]
cmeister2: With regards to this specific change - I think this makes the interface a little nicer if you're only comparing to "instances" of Tags (as per the tests). I'm assuming that Tags are not very big and the general use for this interface is to search for specific Tags.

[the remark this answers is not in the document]
drahnr: The tags are `u32`s in disguise, hence there is no point in taking a reference.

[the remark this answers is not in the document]
drahnr: I'll leave the API as is, since my expectation on a `find` would be to return an `Option`, but this does return a `Result`.

maxdymond (approved):
I can't really comment on the design aspects here - they look reasonable though! The Rust I've already commented on.

cmeister2 (approved):
Oops, wrong account. Looks good :)

(I can't really comment on the design aspects here - they look reasonable though! The Rust I've already commented on.)

dralley:
@drahnr I had asked for "split up tests" to be dropped :(

src/rpm/package.rs
```diff
@@ -142,12 +191,10 @@ impl RPMPackage {
     ///
     ///
     #[cfg(feature = "signature-meta")]
-    pub fn verify_signature<V>(&self, verifier: V) -> Result<(), RPMError>
+    pub fn verify_signature<V>(&self, verifier: V) -> Result<SignatureVerificationOutcome, RPMError>
```
dralley, line 194: @drahnr I'm confused by this. Nothing actually returns `SignatureVerificationOutcome::Fail` so it doesn't seem useful? Even if it did, encoding an error condition in the success path feels very wrong.

src/rpm/package.rs
```diff
@@ -175,7 +222,25 @@ impl RPMPackage {
...
+            &mut header.as_slice(),
+            SeqCursor::new(&[header.as_slice(), self.content.as_slice()]),
+        )?;
+
+        let package_contained = self.metadata.get_digests()?;
+        if recreated_from_content == package_contained {
+            Ok(DigestVerificationOutcome::Match)
+        } else {
+            Ok(DigestVerificationOutcome::Mismatch)
+        }
```
dralley, line 243: A test for this would be good.

Like above though I'm confused about the error strategy.  If I were to write some fairly intuitive code such as

```Rust
pkg.verify_digest()?;
```

it effectively won't actually be checking the digests, and it would be a silent footgun unless the return value was marked as `#[must_use]` and the user checked the value themselves.

But I'm not sure I understand why this is better than `Result<(), RpmError>` with the addition of new error variants for digest mismatch, as we already have one for signature mismatch?

src/rpm/builder.rs
```diff
@@ -537,10 +541,15 @@ impl RPMBuilder {
...
-        let header_and_content_len = header.len() + content.len();
+        let header_and_content_cursor = SeqCursor::new(&[header.as_slice(), content.as_slice()]);
+        let header_and_content_len = header_and_content_cursor.len();
+        let Digests {
+            header_and_content_digest: header_and_content_digest_md5,
+            header_digest: header_digest_sha1,
+        } = RPMPackage::create_digests_from_readers(
+            &mut header.as_slice(),
+            header_and_content_cursor,
+        )?;
```
dralley, line 552: PAYLOADDIGEST and PAYLOADDIGESTALT, which I'm currently adding support for, can't be generated here, because they are put in the main header rather than the signature header.  So I'll rename this to signify that it's only the digests going in the signature header that are created here.

src/rpm/package.rs
drahnr: That's correct, fix eta today

src/rpm/package.rs
drahnr: Rationale was, that the RPM format is fine, the signature just failed to verify. And the verfication might simply caused by a missing key.

But I think for the common usage pattern you're correct that it poses a footgun.

dralley: `VerificationError` could have subtypes potentially

dralley: Actually there is `KeyNotFound` error already https://github.com/rpm-rs/rpm/pull/122/files#diff-6ff19a1ff611f3835b5a025663cf4dec901ed914626f78847414836439bd1a79L77

PR #115, opened 2023-04-08
Clean up and expand tests

dralley:
### 📜 Checklist


test_assets/fixture_packages/SPECS/rpm-feature-coverage.spec
```diff
@@ -0,0 +1,77 @@
+Name:           rpm-feature-coverage
```
dralley, line 1: @DemiMarie I would like to build some nice test RPMs to for use (which you are also welcome to use), do you have any recommendations of other things to exercise?  e.g.

* symlinks
* patches
* scriptlets
* triggers

Or overall feedback, or if you know of a good set of existing ones that won't cause licensing issues.

DemiMarie: Some of the things that should be rejected during parsing:

- `INT32_MIN` trailer offset
- Dribbles in the main header
- Immutable region is not self-consistent
- Wrong amount of padding (too large or too small)
- Misaligned header tag data entry
- Header tag data entry with count 0
- Header tag data entry count too large
- Tag appears in both signature and main headers
- Too many header tag data entries
- Header too large
- Tag that should only be in the signature header is found in the main header (this was an exploit in the past)
- Header-only signature but neither a header+payload signature nor a payload digest (another exploit)
- Invalid tag (such as the one used by RPM internally for public keys) in the signature or main header
- Tag data entry of wrong type

Some things that I reject but librpm does not:

- Dribbles in the signature header.
- Header tags not sorted
- Padding not zeroed
- i18nstring tag has a length more than the i18ntable
- Weak hash algorithms
- Invalid UTF-8 in a string tag
- Suspicious OpenPGP signature
- Non-ASCII string in the signature header

Things that `rpm-oxide` should reject:

- Inconsistent fsverity or IMA signatures (invalid base64, largest one too big, etc)

A simple way to implement this would be to use rpm-oxide for parsing; its API should be able to handle all of this.

dralley: I was thinking more about the happy path at the moment.  Basically - assuming the file is well constructed, how should I exercise as much surface area as possible in terms of parsing and then querying information about the package.

It will make sense to test invalid packages too but it's not my immediate focus.

test_assets/fixture_packages/build_packages.sh
```diff
@@ -1,8 +1,20 @@
...
 rm -r ./BUILD/
+
+RSA_KEY_OPTS=( --define "_gpg_name rsa3072-rpm-signing-key" )
+EDDSA_KEY_OPTS=( --define "_gpg_name ed25519-rpm-signing-key" )
+
+rpmsign --addsign RPMS/x86_64/rpm-empty-0-0.x86_64.rpm $"${EDDSA_KEY_OPTS[@]}"
+rpmsign --addsign RPMS/x86_64/rpm-feature-coverage-2.3.4-5.el8.x86_64.rpm $"${RSA_KEY_OPTS[@]}"
+
+# @todo: how to make IMA signing work? I get errors.
+# --define "_file_signing_key rsa3072-rpm-signing-key" --signfiles
```
dralley, line 20: @puiterwijk What additional steps are needed to make IMA signing work?  Does it need the keys to be in a different format?

test_assets/fixture_packages/SPECS/rpm-feature-coverage.spec
DemiMarie: That makes sense.  `rpm-oxide` was initially written to make sure that invalid packages cannot exploit a potentially-vulnerable version of RPM, so rejecting invalid packages was the single highest priority.  As a result, it follows the “when in doubt, error out” approach, with large-scale testing on entire repositories used to ensure that it really does accept valid packages.  On the other hand, rpm-oxide does not implement package building or installation, as that is simply out of scope for it.  From my perspective, it would make sense for rpm-rs to use rpm-oxide for parsing and cryptographic tasks, while implementing a higher-level API itself.  rpm-oxide does require librpm for cryptography but I am happy to make that configurable via a trait.

As far as package installation is concerned, I suspect the three trickiest parts are likely to be Lua scriptlets, `chroot`, and dependency solving, but I have never implemented this code myself so I have no idea.  You would be better off asking the rpm-ostree developers, as they have (or in the case of Lua scriptlets, need to) implement this.  This would also be a very good opportunity for cooperation.

dralley: Well `rpm-rs` doesn't (and might not ever) handle package installation, because that would involve working with the system `rpmdb`, which is a lot more scope given that details may differ between operating systems and across long timeframes.  I think it actually makes a lot of sense to just use bindings to librpm in that case.  

Anyway, to the original question, we don't need to be concerned with e.g. running scriptlets or doing dependency solving (etc.), but we do want to produce and parse packages with scriptlets or %ghost files or %config or %defattr or patches or different kinds of signatures or  all different kinds of dependencies ... (etc.)  So along those lines, any ideas of what to add to the spec file?

src/rpm/headers/header.rs
```diff
@@ -513,66 +513,8 @@ where
     Ok((input, ()))
 }
 
-#[cfg(test)]
-mod tests2 {
```
dralley, line 517: For some reason this was mid-file.  Moved it to the bottom.

src/tests.rs
```diff
@@ -1,128 +1,21 @@
...
-    let public_key_content = std::fs::read(test_public_key_path())?;
-    let verifier = Verifier::load_from_asc_bytes(&public_key_content).unwrap();
-    package
-        .verify_signature(&verifier)
-        .expect("failed to verify signature");
-    Ok(())
-}
-
+// Currently only testing header internal details which we can't do from the integration tests
+// Once we have enough API surface (or a way to test total equality) we can ditch these
```
dralley, line 12: ^^

tests/parsing.rs
```diff
@@ -0,0 +1,265 @@
...
+            },
+            ChangelogEntry {
+                author: "Lucille Bluth <lucille@bluthcompany.com> - 1.1.1-1".to_owned(),
+                timestamp: 1617192000,
+                description: "- It's a banana, Michael. How much could it cost, $10?".to_owned()
+            },
+        ]
+    );
+
+    // assert_eq!(
```
dralley, line 80: TODO: need to finish https://github.com/rpm-rs/rpm/pull/69 before these are really practical to write tests for that use any kind of normal API.

dralley:
@drahnr This is ready for a review pass.  It could be merged as-is but I think I would like to address a few of the things I called out in comments first.  Also open to feedback on the questions I asked

src/tests.rs
```diff
@@ -494,82 +384,9 @@ fn test_rpm_header_base(package: RPMPackage) -> Result<(), Box<dyn std::error::E
...
 fn test_rpm_header() -> Result<(), Box<dyn std::error::Error>> {
-    let rpm_file_path = test_rpm_file_path();
+    let rpm_file_path = rpm_389_ds_file_path();
     let package = RPMPackage::open(rpm_file_path)?;
     test_rpm_header_base(package)
 }
-
-#[cfg(feature = "signature-meta")]
-#[test]
-fn test_region_tag() -> Result<(), Box<dyn std::error::Error>> {
```
dralley, line 556: Moved to header.rs

tests/compat.rs
```diff
@@ -212,72 +200,17 @@ mod pgp {
         })
     }
 
-    #[test]
-    #[serial_test::serial]
```
dralley, line 216: Moved to tests/signatures.rs

.github/workflows/ci.yml
```diff
@@ -86,7 +86,7 @@ jobs:
       - uses: dtolnay/rust-toolchain@master
         with:
           toolchain: ${{ matrix.rust }}
-      - run: cargo test --features test-with-podman,with-file-async-tokio
+      - run: cargo test --features with-file-async-tokio
```
drahnr, line 89: How come we don't have that feature there anymore?

tests/common.rs
```diff
@@ -0,0 +1,46 @@
+#[allow(dead_code)]
```
drahnr, line 1: How is this dead code? Wouldn't it be preferable to have it gated by the cfg / cfg and split into modules?

tests/compat.rs
```diff
@@ -1,44 +1,32 @@
...
-fn cargo_out_dir() -> std::path::PathBuf {
-    cargo_manifest_dir().join("target")
-}
+mod common;
 
-#[cfg(feature = "signature-meta")]
 use signature::{self, Verifying};
 
-#[cfg(feature = "signature-pgp")]
 mod pgp {
```
drahnr, line 10: Why are these removed?

tests/common.rs
dralley: It's not dead code but rustc thinks it is for some reason.  I'm not sure why, because it really clearly isn't.

tests/compat.rs
dralley: If you look in Cargo.toml, the entire test files are flagged on those features, so putting the flags in the code isn't necessary

.github/workflows/ci.yml
dralley: Well, I was thinking that it didn't make sense any longer if the tests were not going to be directly in the source tree.  But that was fairly early on so I'm pretty neutral again, I can add it back.

drahnr:
I just rebased your PR in `bernhard-foo`.

.github/workflows/ci.yml
drahnr: I don't understand. The idea is to have a controlled environment to verify the built packages are recognized as valid rpms by the existing tooling. So I really would like these to remain. It's a form of acceptance test.

dralley: The tests do remain, they're just not flagged on the feature.  I was thinking there was a better way to control whether the tests were run or not but after searching it appeared there wasn't, particularly.

dralley:
@drahnr Could you just `git revert a408803b13` and push that?  It should still be in your local git history

tests/common.rs
drahnr: If you use `common.rs` in multiple tests, each of these will, be compiled as an individual binaries, so each would have to use all functions to not give warnings.

drahnr:
`bernhard-bar`

.github/workflows/ci.yml
drahnr: And what speaks against keeping the feature flags as a unique discriminator that have a host dependency?

dralley: It just feels weird to have a feature that applies only to tests

drahnr: It's a solution to a problem, if you have a more elegant solution, perfect! Otherwise please keep the feature flag

dralley:
@drahnr Apart from the compat tests, which I'll have working shortly, this is ready for review.  Apologies that it's so large.

tests/compat.rs
```diff
@@ -213,66 +213,12 @@ rpm -vv --checksig /out/{rpm_file} 2>&1
...
+        let package = rpm::Package::open(&out_file)?;
 
         package.verify_signature(verifier)?;
 
         Ok(())
     }
-
-    #[test]
-    #[serial_test::serial]
-    fn test_verify_raw_gpg_signature() -> Result<(), Box<dyn std::error::Error>> {
```
dralley, line 227: I didn't feel like this test was adding sufficient value

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "rpm"
-version = "0.13.0"
+version = "0.13.1"
```
drahnr, line 3: If the above breaks anything, we should make sure to also do a minor crate version bump.

src/constants.rs
```diff
@@ -509,30 +509,26 @@ bitflags! {
         const TRIGGERPREIN = 1 << 25;  // %triggerprein dependency
         const KEYRING	= 1 << 26;
         // bit 27 unused
-        const CONFIG	= 1 << 28;
+        const CONFIG	= 1 << 28;    // config() dependency: TODO: wtf is a config dependency?
```
drahnr, line 512: depends on another package providing a specific configuration file, I don't think the TODO is constructive

src/constants.rs
```diff
@@ -570,7 +566,7 @@ bitflags! {
         const DONOTUSE = 1 << 2;  // %%donotuse
         const MISSINGOK = 1 << 3;  // %%config(missingok)
         const NOREPLACE = 1 << 4;  // %%config(noreplace)
-        // const SPECFILE = 1 << 5;  // first file in SRPM?
+        const SPECFILE = 1 << 5;  // specfile, which is the first file in a source RPM
```
drahnr, line 569: :+1:

src/rpm/builder.rs
```diff
@@ -663,11 +663,23 @@ impl PackageBuilder {
         let mut file_verify_flags = Vec::with_capacity(files_len);
         let mut dir_indixes = Vec::with_capacity(files_len);
         let mut base_names = Vec::with_capacity(files_len);
+        let mut users_to_create = HashSet::new();
+        let mut groups_to_create = HashSet::new();
 
         let mut combined_file_sizes: u64 = 0;
+        let mut has_caps = false;
```
drahnr, line 670:
```suggestion
        let mut has_capabilities = false;
```

src/rpm/headers/types.rs
```diff
@@ -332,19 +342,61 @@ impl Dependency {
...
+            "".to_string(),
+        )
+    }
+
+    /// Create a dependency on a package or file required for a pre-un-install script.
+    pub fn script_preun(dep_name: impl Into<String>) -> Self {
+        Self::new(
+            dep_name.into(),
+            DependencyFlags::SCRIPT_PREUN,
+            "".to_string(),
```
drahnr, line 366:
```suggestion
            String::new(),
```

src/rpm/headers/types.rs
```diff
@@ -332,19 +342,61 @@ impl Dependency {
...
+        )
+    }
+
+    /// Create a dependency on a package or file required for a post-un-install script.
+    pub fn script_postun(dep_name: impl Into<String>) -> Self {
+        Self::new(
+            dep_name.into(),
+            DependencyFlags::SCRIPT_POSTUN,
+            "".to_string(),
```
drahnr, line 375:
```suggestion
            String::new(),
```

src/rpm/headers/types.rs
```diff
@@ -332,19 +342,61 @@ impl Dependency {
...
         Self::new(
             dep_name.into(),
-            DependencyFlags::RPMLIB | DependencyFlags::EQUAL,
+            DependencyFlags::RPMLIB | DependencyFlags::LESS | DependencyFlags::EQUAL,
             version.into(),
         )
     }
 
-    fn new(dep_name: String, flags: DependencyFlags, version: String) -> Self {
+    // TODO: wtf is a config dependency?
```
drahnr, line 388:
```suggestion
```

tests/building.rs
```diff
@@ -0,0 +1,272 @@
...
+
+/// Build an empty source package and compare it to one built by rpmbuild
+/// Blocked on completion of source package support
+#[ignore = "https://github.com/rpm-rs/rpm/issues/66"]
+#[test]
+fn test_empty_source_package_equivalent() -> Result<(), Box<dyn std::error::Error>> {
+    let empty_rpm = Package::open(common::rpm_empty_source_path())?;
+    let built_empty_rpm = PackageBuilder::new("rpm-empty", "0", "LGPL", "x86-64", "").build()?;
+
+    // @TODO: currently failing due to missing tags, different checksums, offsets etc.
```
drahnr, line 40:
```suggestion
    // TODO: currently failing due to missing tags, different checksums, offsets etc.
```

tests/building.rs
```diff
@@ -0,0 +1,272 @@
...
+
+/// Build an empty package and compare it to one built by rpmbuild
+#[test]
+fn test_empty_package_equivalent() -> Result<(), Box<dyn std::error::Error>> {
+    let empty_rpm = Package::open(common::rpm_empty_path())?;
+    let built_empty_rpm = PackageBuilder::new("rpm-empty", "0", "LGPL", "x86-64", "")
+        .compression(CompressionType::None)
+        .build()?;
+
+    // @TODO: currently failing due to missing tags, different checksums, offsets etc.
```
drahnr, line 18:
```suggestion
    // TODO: currently failing due to missing tags, different checksums, offsets etc.
```

tests/compat.rs
drahnr: It does verify the signature creation done with `rpmbuild` matches that of `rpm-rs`, which I do find a worthwhile thing to check against.

Could you please re-add it?

tests/compat.rs
```diff
@@ -354,110 +300,33 @@ fn podman_container_launcher(
...
-#gpg --batch --generate-key gpgkeyspec  2>&1
-
-### or (which has a couple of advantages regarding reproducibility)
-
-export PK=/assets/public_key.asc
-export SK=/assets/secret_key.asc
-
-gpg --allow-secret-key-import --import "${SK}" 2>&1
-gpg --import "${PK}" 2>&1
+pushd assets/
```
drahnr, line 303: This is unfortunately not present by default on quite a few linux distributions.

drahnr (changes requested):
A few minor nits, a few small comments and cleanups, the majority looks excellent!

Thank you, and a Merry Christmas :christmas_tree: !

src/constants.rs
dralley: The part that I find confusing is what makes it different from any other "file" dependency, and why not depend on that config file directly rather than `config(rpm-basic)` 1.0-1, and why packages both provide it and also self-require it as seen here.

https://github.com/rpm-rs/rpm/pull/115/files#diff-26e82b76ab9e2c9c188e78b8b38612a335b74fb0295201fbd609a74a45582f93R194

I'd also kinda like to know which line is triggering it, I can figure that out myself but I haven't done so yet.

Cargo.toml
dralley: I should clarify, 0.13.1 is already released but the version hasn't been bumped on main branch yet - I forgot to do that when merging the changelog, so I ended up just adjusting that locally when I did the release.

Prior to the next release, it would be bumped again alongside the changelog.

tests/compat.rs
dralley: I wasn't under the impression that these tests were meant to be totally generic, given that they call `yum`, `dnf`, `rpm`, use podman, etc.  

Is it OK, given that pretty much every distro we would care about (SUSE, Fedora, CentOS/RHEL) has it available?

tests/compat.rs
dralley: That's what the previous test does.  What this test seems to be doing, is directly sign a bit of data with `gpg`, and verify it with `Verifier`.  RPMs aren't involved.

It's not a *bad* thing to test exactly, but it's not providing any additional coverage beyond what the other tests provide, it takes a long time to run, and it's not a simple test.  If we were to keep it I'd rather just commit a signature of one of the existing fixture files as a fixture.

src/constants.rs
dralley: @DemiMarie Do you happen to know what is the intended purpose behind putting `config()` dependencies in both the provides and requires sections?

tests/compat.rs
drahnr: It adds a barrier, `rpm` can be reasonably expected, I i.e. don't have these installed on any of my machines (fedora 39).

dralley: `pushd` and `popd` are bash builtins, they aren't installed separately, but they're available on any distro that uses bash as the shell, which includes Fedora.

```
dalley@thinkpad:~/Devel/rpm-dev/rpm-rs/tests/assets$ which pushd
/usr/bin/which: no pushd in (/home/dalley/.cargo/bin:/home/dalley/.local/bin:/home/dalley/bin:/usr/share/Modules/bin:/usr/local/bin:/usr/local/sbin:/usr/bin:/usr/sbin)
dalley@thinkpad:~/Devel/rpm-dev/rpm-rs/tests/assets$ pushd RPMS
~/Devel/rpm-dev/rpm-rs/tests/assets/RPMS ~/Devel/rpm-dev/rpm-rs/tests/assets
dalley@thinkpad:~/Devel/rpm-dev/rpm-rs/tests/assets/RPMS$ popd
~/Devel/rpm-dev/rpm-rs/tests/assets

```

tests/assets/SPECS/rpm-file-attrs.spec
```diff
@@ -0,0 +1,68 @@
...
+mkdir -p %{buildroot}/opt/%{name}
+
+echo "normal" > %{buildroot}/opt/%{name}/normal
+mkdir -p %{buildroot}/opt/%{name}/dir/
+
+# symlinks to a file and directory
+touch %{buildroot}/opt/%{name}/symlink
+mkdir -p %{buildroot}/opt/%{name}/symlink_dir/
+ln -sf ../normal %{buildroot}/opt/%{name}/symlink
+ln -sf ../dir/ %{buildroot}/opt/%{name}/symlink_dir/
```
dralley, line 21: This is also really confusing.  The only way I was able to create a symlink and get the package to build without a warning, was to touch the files first, and then create the link with overwriting enabled.  

I strongly suspect I'm doing something wrong here.

tests/assets/SPECS/rpm-file-attrs.spec
```diff
@@ -0,0 +1,68 @@
...
+mkdir -p %{buildroot}/opt/%{name}
+
+echo "normal" > %{buildroot}/opt/%{name}/normal
+mkdir -p %{buildroot}/opt/%{name}/dir/
+echo "file-in-a-dir" > %{buildroot}/opt/%{name}/dir/normal
+
+# symlinks to a file and directory
+mkdir -p %{buildroot}/opt/%{name}/symlink_dir/
+ln -s ./normal %{buildroot}/opt/%{name}/symlink
+ln -s ../dir %{buildroot}/opt/%{name}/symlink_dir/
```
dralley, line 21: To add a bit more specificity to what was going wrong earlier,

without the "mkdir" it yields an error "No such file or directory" if there is a trailing slash, or "Not a directory" if there is not.  With the "mkdir" it succeeds but the symlink is placed inside of "symlink_dir" instead of at that path.

I still find this confusing, because when I create that symlink manually without "mkdir" it works fine.  Only inside of rpm does it seem to fail.

dralley:
I've cleaned up the commits a bit and made some adjustments to the specfiles.  There are a few other simple cleanups as well but nothing significant.

.github/workflows/ci.yml
```diff
@@ -63,6 +62,11 @@ jobs:
         flags:
           - "--all-features"
           - "--no-default-features"
+        os:
+          - "ubuntu-latest"
+          - "macos-latest"
+          # - "windows-latest" TODO
```
dralley, line 68: Filed an issue for this.

tests/compat.rs
drahnr: It doesn't sign an _rpm_, but it does call the signature process, and hence verifies the signing command of `rpmbuild` being consistent with the signing of `rpm-rs`.

It adds value, such that it tests exactly one thing, rather than the entire process. To improve the runtime, we should always compile the cryptographic dependency libraries with optimizations enabled.

tests/compat.rs
drahnr: You are correct, as a `zsh` user I wasn't aware.

tests/compat.rs
dralley: >It doesn't sign an rpm, but it does call the signature process, and hence verifies the signing command of rpmbuild being consistent with the signing of rpm-rs.

At no point in this test is the signing of `rpm-rs` being tested or compared against `rpmbuild`.  `Verifier` is exercised, but `Signer` is not.  One single signature is created using a shell script using GPG which we provide, which isn't necessarily consistent with either `rpm-rs` or `rpm`, and then verified.  

To do what you're discussing, you'd want to sign both ways and then directly compare the two signatures, but that isn't happening here.

tests/assets/SPECS/rpm-file-attrs.spec
drahnr: Could you add that as an information nugget, it will be lost if it's only in the PR

src/rpm/headers/types.rs
```diff
@@ -53,7 +53,50 @@ pub const REGULAR_FILE_TYPE: u16 = 0o100000; //  bit representation = "100000000
...
+
+// typedef enum rpmFileTypes_e {
+//     	=  1,	/*!< pipe/fifo */
+//     CDEV	=  2,	/*!< character device */
+//     XDIR	=  4,	/*!< directory */
+//     BDEV	=  6,	/*!< block device */
+//     REG		=  8,	/*!< regular file */
+//     LINK	= 10,	/*!< hard link */
+//     SOCK	= 12	/*!< socket */
+// } rpmFileTypes;
```
drahnr, line 66: Can we represent this as another bitfield?

dralley:
Oh, I just rebased this PR, it's not ready for merging yet or anything.

dralley:
Also needs tests for:

* package iteration, particularly `files()`, for both cpio and stripped payload formats
* generation of packages with cpio and stripped payload formats
* package extraction
* multiple package signatures

dralley:
This is not a perfect PR, but I'm merging it to unlock other work from going forwards, and the tests can be fixed / expanded later (prior to the next release)

PR #116, opened 2023-04-08
Simplify Tag trait

dralley:
There's no need for Tag to be a blanket impl when we know that there are only 2 implementers, and TagName isn't useful when you can just implement it on Tag.

### DIFF
--- a/src/constants.rs
+++ b/src/constants.rs
@@ -423,18 +423,24 @@ pub enum IndexSignatureTag {
     RPMTAG_INSTALLTIME = IndexTag::RPMTAG_INSTALLTIME as u32,
 }
 
-pub trait TypeName {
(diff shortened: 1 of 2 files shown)


src/constants.rs
```diff
@@ -423,17 +423,23 @@ pub enum IndexSignatureTag {
     RPMTAG_INSTALLTIME = IndexTag::RPMTAG_INSTALLTIME as u32,
 }
 
-pub trait TypeName {
```
dralley, line 426: `TypeName` is only used for `Tag`, and there are only 2 implementers of `Tag`, so we can ditch the blanket impl and `TypeName` trait.

cmeister2 (approved):
LGTM

PR #118, opened 2023-04-08
Automatically set BUILDHOST and tag and add COOKIE tag

dralley:
Incremental improvement towards making an empty package built by rpm-r more similar to one created by rpmbuild

### 📜 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `$pkg.metadata.get_changelog_entries()`
 - Added the following functions to `RPMBuilder` to support weak dependencies: `recommends()`,
   `suggests()`, `enhances()` and `supplements()`
+- Added the following additional functions to `RPMBuilder`: `cookie()`, `build_host()`
+- Added `get_build_cookie()` for use with `RPMBuilder::cookie()`
 - Added the following functions to `$pkg.metadata` for retrieval of various kinds of RPM
   dependencies: `get_provides()`, `get_requires()`, `get_obsoletes()`, `get_conflicts()`,
   `get_recommends()`, `get_suggests()`, `get_enhances()`, `get_supplements()`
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -32,7 +34,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Breaking Changes
 
-- Bump MSRV to 1.63.0
+- Bump MSRV to 1.64.0
```
cmeister2, line 37: It'd be nice if we had an idea of what was pinning the MSRV high. It's better if the MSRV is lower, cause then implementors can have a lower MSRV and it's compatible with more versions of Rust.

src/rpm/builder.rs
```diff
@@ -67,17 +69,31 @@ async fn async_file_mode(_file: &async_std::fs::File) -> Result<u32, RPMError> {
...
 
+/// Returns a value that can be used for associating several package builds as being part of one operation
+///
+/// You can use any value, but the standard format is "${build_host} ${build_time}"
+///
+/// See `RPMBuilder::cookie()`
+pub fn get_build_cookie() -> String {
+    let build_time = SystemTime::now()
+        .duration_since(SystemTime::UNIX_EPOCH)
+        .unwrap()
```
cmeister2, line 80: Is this an infallible operation? Might be nice to be `expect()` if so.

CHANGELOG.md
dralley: In this case `gethostname` wanted an MSRV bump.  With an older version of `gethosname` it could possibly be pushed lower.

src/rpm/builder.rs
dralley: It ought to be effectively infalliable since it only returns `Err` when "earlier is later than self".  So the system timestamp would have to be less than 0.

dralley:
@cmeister2 Done

cmeister2 (approved):

drahnr (commented):
@cmeister2 @dralley 

The changes in this PR are yepordizing binary reproducibility for sake of identical behavior to existing rpm tooling, which was never a goal for `rpm-rs`. I don't oppose the _possibility_ of providing such data - adding timestamps and build host info should, at most, be an explicit OPT-IN as part of the builder, and not the default in any case.

dralley:
@drahnr See the 2nd-to-last commit on https://github.com/rpm-rs/rpm/pull/115

Issue #117, opened 2023-04-08
Ability to create reproducable builds


dralley:
We need to have the  same knobs that rpmbuild has to enable reproducible package builds.  Timestamps and various other things get in the way.

As far as I know the needed features are (equivalents to):

> \# Have SOURCE_DATE_EPOCH value set
> %source_date_epoch_from_changelog 1
> \# Ensure no file timestamps are later than SOURCE_DATE_EPOCH
> %clamp_mtime_to_source_date_epoch 1
> \# Force build timestamp to SOURCE_DATE_EPOCH
> %use_source_date_epoch_as_buildtime 1
> \# Make the hostname deterministic
> %_buildhost mockbuild

SOURCE_DATE_EPOCH is referring to https://reproducible-builds.org/docs/source-date-epoch/

dralley:
SOURCE_DATE_EPOCH support is now implemented but in a slightly different way from rpmbuild.  If the environment variable is set and parses correctly, we just use it, whereas rpmbuild has a separate flag for enabling it that you have to configure.

We don't currently have a way to clamp file mtimes, nor use the most recent changelog timestamp as the build timestamp.  Maybe it makes sense to put all of this behind one "reproducable" option on `RPMBuilder`?

Open for discussion.

drahnr:
I think the default should be to build reproducible rpms, the user then can opt in to provide i.e. the current timestamp or file timestamps or buildhost they desire. This would also enable the option to shed a few dependencies for i.e. the `buildhost` since the user would have to provide it, we don't need to have a dependency on the `gethostname` crate.

dralley:
I'm told that in practice those fields are quite useful for debugging purposes, so it's actually useful to opt-in to reproducibility rather than the other way around.

dralley:
However, I can presume that is with regards to `rpmbuild`.  Programmatically generating RPMs is slightly different.  I'm legitimately open to discussion on the default and how it's configured

dralley:
If you want to take this issue and https://github.com/rpm-rs/rpm/pull/115/commits/5cfc3be4c67b20f259cce30f41f3a6a057e70b57#diff-b82882f602be2cf2d09fd6546c2097235d9417da83223e9516b8eb783c6257ffR233 and work on it, then that is fine with me.

dralley:
@drahnr How do you think

># Ensure no file timestamps are later than SOURCE_DATE_EPOCH
> %clamp_mtime_to_source_date_epoch 1

[...]

PR #121, opened 2023-04-15
refactor: opt-in for non-reproducible builder options

drahnr:
Makes sure `build_time` and `build_host` are opt-in.

Not clear, if there is a reason to keep `cookie`. We're a lib, and as such that's more a design choice for the user of the library than us.

Ref #117


src/rpm/compressor.rs
```diff
@@ -7,6 +7,22 @@ pub enum Compressor {
...
+        Self::None(Default::default())
+    }
+}
+
+impl std::fmt::Debug for Compressor {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str(match self {
+            Self::None(_) => "Uncompressed",
+            Self::Gzip(_) => "compressed/gzip",
+            Self::Zstd(_) => "comrpessed/zstd",
```
cmeister2, line 21: Sp: compressed

src/rpm/builder.rs
```diff
@@ -148,32 +130,7 @@ impl RPMBuilder {
...
-            files: BTreeMap::new(),
-            changelog_authors: Vec::new(),
-            changelog_entries: Vec::new(),
-            changelog_times: Vec::new(),
-            compressor: Compressor::None(Vec::new()),
-            directories: BTreeSet::new(),
-            vendor: None,
-            url: None,
-            vcs: None,
+            ..Default::default()
```
cmeister2, line 133: I could see a world where the replaced functionality is a feature flag. Have `hostname` as an optional package, and have a feature flag for "rpmbuild" or something, then set the values for the build host and build time conditionally if the feature flag is enabled. All it would be doing is setting the options to the calculated values.

cmeister2 (commented):
Couple of thoughts

src/rpm/builder.rs
drahnr: But why would we pull in dependencies on host functionality if we can avoid it and just expect the user to feed whatever they desire? I'd prefer to keep the feature space narrow and push things out as much as we can. Adding to many feature flags is a footgun/slowdown for development, so I'd prefer to keep it outside of this package/repo/crate. If the user wants to, they can still pull whatever dependencies in.

src/lib.rs
```diff
@@ -40,8 +40,10 @@
 //!                         .user("hugo"),
 //!             )?
 //!             .pre_install_script("echo preinst")
-//!             .add_changelog_entry("me", "was awesome, eh?", 123123123)
-//!             .add_changelog_entry("you", "yeah, it was", 12312312)
+//!             .build_time(std::time::SystemTime::now())
+//!             .build_host(gethostname::gethostname().to_str().unwrap().to_string())
+//!             .add_changelog_entry("me", "was awesome, eh?", std::time::SystemTime::now())
```
dralley, line 45: I don't know that we should encourage using it like this, changelog times ought to be static and you wouldn't want multiple of them at the same timestamp.

CHANGELOG.md
```diff
@@ -33,6 +33,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   did not have any files associated.
 - Ensured that digests are always added to built RPMs. Previously they would not be included unless
   the "signature-meta" (or "signature-pgp") features were enabled.
+- Move to opt-in for reproducability inhibiting by default values, `build_time` and `build_host`.
+- Removed `get_build_cookie` again
```
dralley, line 37: I think we should keep it personally.  It's so simple there's limited reason not to, and makes it easy for users to do the standard thing even if they don't know exactly what the "standard thing" is, which is useful.

But for this or anything else it it would be better to just remove the changelog line where it was added.

src/rpm/compressor.rs
```diff
@@ -7,6 +7,22 @@ pub enum Compressor {
     Zstd(zstd::stream::Encoder<'static, Vec<u8>>),
 }
 
+impl Default for Compressor {
+    fn default() -> Self {
+        Self::None(Default::default())
```
dralley, line 12: I kinda feel like no compression is a bad default?  Gzip is the default for `rpmbuild` and is supported by everything, Zstd is much better and is supported by everything for the past 5-6 years (e.g. EL8)

src/tests/fut.rs
```diff
@@ -37,8 +37,8 @@ async fn test_rpm_builder_async() -> Result<(), Box<dyn std::error::Error>> {
         )
         .await?
         .pre_install_script("echo preinst")
-        .add_changelog_entry("me", "was awesome, eh?", 123123123)
-        .add_changelog_entry("you", "yeah, it was", 12312312)
+        .add_changelog_entry("me", "was awesome, eh?", std::time::SystemTime::now())
+        .add_changelog_entry("you", "yeah, it was", std::time::SystemTime::now())
```
dralley, line 41: This is why I wanted to go ahead and revert those changes, so it doesn't continually cause rebase issues.

src/rpm/builder.rs
dralley: If you look at the implementation, the `gethostname` dependency is only about 30 lines of code

src/rpm/builder.rs
```diff
@@ -196,18 +153,20 @@ impl RPMBuilder {
...
 
-    /// Override the automatically-detected name of the build host with a new value
-    pub fn build_host(mut self, build_host: &str) -> Self {
-        self.build_host = build_host.to_owned();
+    pub fn build_host(mut self, build_host: impl AsRef<str>) -> Self {
+        self.build_host = Some(build_host.as_ref().to_owned());
+        self
+    }
+
+    pub fn build_time(mut self, build_time: SystemTime) -> Self {
```
dralley, line 161: Honestly I'm still partial to doing something like this: https://github.com/rpm-rs/rpm/commit/5cfc3be4c67b20f259cce30f41f3a6a057e70b57#diff-b82882f602be2cf2d09fd6546c2097235d9417da83223e9516b8eb783c6257ffR229-R232

We ought to support `SOURCE_DATE_EPOCH`, it's not an RPM specific thing, it's a cross-distro standard for reproducible builds.

https://reproducible-builds.org/docs/source-date-epoch/

src/rpm/builder.rs
```diff
@@ -216,13 +175,18 @@ impl RPMBuilder {
         self
     }
 
-    pub fn add_changelog_entry<E, F>(mut self, author: E, entry: F, time: u32) -> Self
+    pub fn add_changelog_entry<A, E>(
+        mut self,
+        author: A,
+        entry: E,
+        time: std::time::SystemTime,
```
dralley, line 182: I don't know about this, Unix timestamps are more stable because they never touch timezones.  If a user stores their time as a unix timestamp (like in a database), now there's an unnecessary round trip.  A vec of `SystemTime` would also require 4x as many bits as a vec of u32.

CHANGELOG.md
drahnr: It adds the `gethostname` dependency and I see this as an Impl detail of the user of the library.

src/rpm/compressor.rs
drahnr: It's primarily for builder use, where that was the default.

src/rpm/builder.rs
drahnr: This is out of scope for the library. It's not a tool, it's a building block/lib that may be used for tools. Providing API to adhere to such specs sure, but not implement any recommendation that's out there.

src/lib.rs
drahnr: I just converted the example I think?

src/tests/fut.rs
drahnr: Let's merge your test refactor first and get that out of the way

src/lib.rs
dralley: Exactly, we're doing so in an example - that's encouraging :)

src/rpm/builder.rs
drahnr: I think that's our main cause of disagreement. I vigorously want to limit scope, enable but not cater Implementation of best practices.

src/rpm/builder.rs
drahnr: It's only ever stored in the builder, and I prefer clear types over floating integers that are just bound by documentation. I'd prefer to define a trait `IntoUnixTimeOffset` and `struct UnixTimeOffset` and make it a traitbound on the arg and convertible from both, `SystemTime` and `u32`

src/rpm/builder.rs
drahnr: I always ask myself: "Is this our core concern?" "Are we adding deps that many users would possibly not desire having or prevents them from using?" And to me it's a clear 1) no - only for a vague debugging and origin tracking which is highly context specific 
2) likely no, get hostname is not supported on all platforms at least from memory

src/rpm/compressor.rs
dralley: I get that it was the default, but is it a good default?  (we can address that separately, not asking for it in this PR)

drahnr: I agree with your statement. Parsing is not affected, it's just the identical default for constructing the builder

dralley: My background is software for managing RPMs, and our users sometimes have several terabytes of RPMs, so this is the kind of thing I think about.

src/rpm/builder.rs
dralley: Looking at https://doc.rust-lang.org/std/time/struct.SystemTime.html

The only constructor I can see is `SystemTime::now()`  and I don't see any options for `From<..>`.  That's an issue

dralley: Considering it ends up as a Vec<u32> anyway, I don't really see the point of storing it as something else

src/tests/fut.rs
dralley: I'll split the PR in half so we can merge the refactors while I'm still working on the rewrites.

dralley:
@drahnr Let's do a new release after we finish up + merge this PR and the payloaddigest one.

src/rpm/compressor.rs
dralley: Created a discussion here: https://github.com/rpm-rs/rpm/discussions/127

src/rpm/builder.rs
dralley: For what it's worth, I'm pretty sure rpmlib when used as a library still respects this.  However It's not a hill I will die on

src/rpm/builder.rs
dralley: Well it has implementations for (Li|U)nix and Windows, so that covers most platforms, but fair enough.  I understand your PoV.

src/compat_tests.rs
```diff
@@ -183,8 +192,16 @@ mod pgp {
             )?
             .epoch(1)
             .pre_install_script("echo preinst")
-            .add_changelog_entry("me", "was awesome, eh?", 123_123_123)
-            .add_changelog_entry("you", "yeah, it was", 12_312_312)
+            .add_changelog_entry(
+                "me",
+                "was awesome, eh?",
+                chrono::Utc.timestamp_millis_opt(1681411811).unwrap(),
```
dralley, line 198: I think you want `timestamp_opt()` not `timestamp_millis_opt()`

CHANGELOG.md
dralley: Ok.  Well, the second points remains, just alter / remove the original changelog entry rather than adding a new one.

src/compat_tests.rs
drahnr: Good catch!

src/rpm/builder.rs
```diff
@@ -68,24 +65,21 @@ async fn async_file_mode(_file: &async_std::fs::File) -> Result<u32, RPMError> {
     Ok(0)
 }
 
-/// Returns a value that can be used for associating several package builds as being part of one operation
-///
-/// You can use any value, but the standard format is "${build_host} ${build_time}"
-///
-/// See `RPMBuilder::cookie()`
```
dralley, line 75: Move this description to `RPMBuilder::cookie()` please.

https://github.com/rpm-rs/rpm/blob/815ddb38afef051c2e3dfb0cf8eae99f5c96aa32/src/rpm/builder.rs#L205-L212

src/rpm/builder.rs
```diff
@@ -196,18 +157,20 @@ impl RPMBuilder {
         self
     }
 
-    /// Override the automatically-detected name of the build host with a new value
-    pub fn build_host(mut self, build_host: &str) -> Self {
-        self.build_host = build_host.to_owned();
+    pub fn build_host(mut self, build_host: impl AsRef<str>) -> Self {
```
dralley, line 160: Could also you add docstrings to `build_host`, `build_time`, and `add_changelog` entry with your examples

CHANGELOG.md
```diff
@@ -33,6 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   did not have any files associated.
 - Ensured that digests are always added to built RPMs. Previously they would not be included unless
   the "signature-meta" (or "signature-pgp") features were enabled.
+- Move to opt-in for reproducability inhibiting by default values, `build_time` and `build_host`.
```
dralley, line 35: The way this reads makes it kind of ambiguous, "reproducability inhibiting" and "by default" need hyphens.  But rewriting it might be better:

```suggestion
- To facilitate reproducible builds, stop writing `build_time` to the package by default.  Users can configure it with `RPMBuilder::build_time()`.
```

We've never had a release with `build_host` so no point mentioning that.

src/rpm/builder.rs
```diff
@@ -216,14 +202,21 @@ impl RPMBuilder {
         self
     }
 
-    pub fn add_changelog_entry<E, F>(mut self, author: E, entry: F, time: u32) -> Self
+    pub fn add_changelog_entry<A, E, TZ>(
```
dralley, line 205: Still missing a documentation string

tests/compat.rs
```diff
@@ -160,8 +169,16 @@ mod pgp {
             )?
             .epoch(1)
             .pre_install_script("echo preinst")
-            .add_changelog_entry("me", "was awesome, eh?", 123_123_123)
-            .add_changelog_entry("you", "yeah, it was", 12_312_312)
+            .add_changelog_entry(
+                "me",
```
dralley, line 173: We should use the standard packaging guidelines for the author value, which means

`FirstName LastName <email> - VersionAndRelease`

The description should start with `-`

So for the changelog entry (in a specfile)

> %changelog
> \* Tue May 31 2016 Adam Miller <maxamillion@fedoraproject.org> - 0.1-1
> \- First bello package
> \- Example second

The date would go to date, "Adam Miller <maxamillion@fedoraproject.org> - 0.1-1" would go to "author", and 

> \- First bello package
> \- example second item in the changelog for version-release 0.1-1

would be the description

dralley:
@drahnr After that ^^ this is good to merge

tests/compat.rs
drahnr: I think that needs some more refinement, why would we make the version suffix `- 0.1-1` part of the author? It appears the API is not quite there and I'd like to carve that out in a separate PR.

drahnr:
> @drahnr After that ^^ this is good to merge

Given the remaining questions, can we push that to a follow up PR?

dralley:
I don't think we need to change the API.  It's weird, I agree, but the fact is that the data is actually stored together in the same tag, and other libraries like createrepo_c keep them together in the API.

What you could do is change `author` to `changelogname` or something less specific.  That way it's at least not as confusing.  With maybe a bit of documentation to go along with it.

src/rpm/builder.rs
```diff
@@ -217,14 +201,38 @@ impl RPMBuilder {
...
+    ///     .add_changelog_entry(
+    ///         "Gambl B. Xen <gbx@example.com>",
+    ///         " - Add enumerator."
+    ///         rpm::chrono::DateTime::parse_from_rfc3339("1996-12-19T16:39:57-08:00").unwrap(),
+    ///     )
+    ///     //..
+    /// ```
+    pub fn add_changelog_entry<A, E, TZ>(
+        mut self,
+        author: A,
```
dralley, line 223: e.g. you could say something along the lines of

> Convention is to have one changelog entry per release.  That the changelog name should be comprised of the first and last name of the release publisher, followed by their email address inside angle brackets, followed by a hyphen separator, followed by the version identifier of the release.  Changes within the release should be listed in a bulleted list, one-per-line, using a hyphen character as a bullet.

dralley: And rename the parameter

```suggestion
        changelogname: N,
```

drahnr:
The entire API here is just quirkesque. I'll create a separate PR to straighten this with a separate type that exposes a proper builder pattern.

dralley:
I don't think it's terribly quirky tbh, it's in line with what any other library does.

drahnr:
> I don't think it's terribly quirky tbh, it's in line with what any other library does.

It's not idiomatic Rust, it's not intuitive for anyone using it the first time and issues are only encountered with other libraries validating the changelog format. That is bad UX.

dralley:
Failing test

drahnr:
> Failing test

Fixed

dralley (approved):

PR #123, opened 2023-04-15
Refactor / clean up tests

dralley:
The first half of the other PR, without new tests or rewriting any of them.


tests/compat.rs
```diff
@@ -1,85 +1,62 @@
...
     use super::*;
-    use futures::io::AsyncWriteExt;
     use signature::pgp::{Signer, Verifier};
-    use tokio_util::compat::TokioAsyncReadCompatExt;
+    use std::str::FromStr;
 
-    #[tokio::test]
     #[serial_test::serial]
-    async fn create_full_rpm_async() -> Result<(), Box<dyn std::error::Error>> {
+    fn create_full_rpm() -> Result<(), Box<dyn std::error::Error>> {
```
dralley, line 17: We test async building here already, so we may as well simplify the feature matrix for these tests: https://github.com/rpm-rs/rpm/pull/123/files#diff-a0262fea108af44a98cc49e5eb72641963dd816513f2d0a22eb738fcfed55ebeR303

Later we can just compare the output of async to not-async building, and that will give us coverage of async automatically, presuming we keep async around.

tests/common.rs
```diff
@@ -0,0 +1,46 @@
+#[allow(dead_code)]
```
drahnr, line 1:
```suggestion
#![allow(dead_code)]
```

drahnr (approved):

PR #124, opened 2023-04-15
Create & verify PAYLOADDIGEST, PAYLOADDIGESTALT

dralley:
### 📜 Checklist

### DIFF
--- a/src/rpm/builder.rs
+++ b/src/rpm/builder.rs
@@ -8,6 +8,7 @@ use std::os::unix::fs::PermissionsExt;
 use std::path::{Path, PathBuf};
 use std::time::{SystemTime, UNIX_EPOCH};
 
+use digest::Digest;
 use gethostname::gethostname;
 
 use super::compressor::Compressor;
@@ -328,8 +329,6 @@ impl RPMBuilder {
         modified_at: u32,
         options: RPMFileOptions,
     ) -> Result<(), RPMError> {
-        use sha2::Digest;
-
         let dest = options.destination;
         if !dest.starts_with("./") && !dest.starts_with('/') {
             return Err(RPMError::InvalidDestinationPath {
@@ -592,6 +591,7 @@ impl RPMBuilder {
         // Lead is not important. just build it here
 
         let lead = Lead::new(&self.name);
+        let mut archive: Vec<u8> = Vec::new();
 
         let mut ino_index = 1;
 
@@ -645,13 +645,14 @@ impl RPMBuilder {
                 .ino(ino_index)
                 .uid(self.uid.unwrap_or(0))
                 .gid(self.gid.unwrap_or(0))
-                .write(&mut self.compressor, content.len() as u32);
+                .write(&mut archive, content.len() as u32);
 
             writer.write_all(&content)?;
             writer.finish()?;
 
             ino_index += 1;
         }
+        cpio::newc::trailer(&mut archive)?;
 
         self.provides
             .push(Dependency::eq(self.name.clone(), self.version.clone()));
@@ -962,6 +963,41 @@ impl RPMBuilder {
 
         let possible_compression_details = self.compressor.get_details();
 
+        self.compressor.write_all(&archive)?;
+        let payload = self.compressor.finish_compression()?;
+
+        // digest of the uncompressed raw archive
+        let raw_payload_digest_sha256 = {
+            let mut hasher = sha2::Sha256::default();
+            hasher.update(archive.as_slice());
+            hex::encode(hasher.finalize())
+        };
+
+        // digest of the compressed archive (payload)
+        let payload_digest_sha256 = {
+            let mut hasher = sha2::Sha256::default();
+            hasher.update(payload.as_slice());
+            hex::encode(hasher.finalize())
+        };
+
+        actual_records.extend([
+            IndexEntry::new(
+                IndexTag::RPMTAG_PAYLOADDIGEST,
+                offset,
+                IndexData::StringArray(vec![payload_digest_sha256]),
+            ),
+            IndexEntry::new(
+                IndexTag::RPMTAG_PAYLOADDIGESTALGO,
+                offset,
+                IndexData::Int32(vec![FileDigestAlgorithm::Sha2_256 as u32]),
+            ),
+            IndexEntry::new(
+                IndexTag::RPMTAG_PAYLOADDIGESTALT,
+                offset,
+                IndexData::StringArray(vec![raw_payload_digest_sha256]),
+            ),
+        ]);
+
         if let Some(details) = possible_compression_details {
             actual_records.push(IndexEntry::new(
                 IndexTag::RPMTAG_PAYLOADCOMPRESSOR,
@@ -1184,9 +1220,7 @@ impl RPMBuilder {
         }
 
         let header = Header::from_entries(actual_records, IndexTag::RPMTAG_HEADERIMMUTABLE);
-        self.compressor = cpio::newc::trailer(self.compressor)?;
-        let content = self.compressor.finish_compression()?;
 
-        Ok((lead, header, content))
+        Ok((lead, header, payload))
     }
 }
--- a/src/rpm/builder.rs
+++ b/src/rpm/builder.rs
@@ -591,7 +591,13 @@ impl RPMBuilder {
         // Lead is not important. just build it here
 
         let lead = Lead::new(&self.name);
-        let mut archive: Vec<u8> = Vec::new();
+        let possible_compression_details = self.compressor.get_details();
+
+        // Make the borrow checker happy
+        let is_zstd = matches!(self.compressor, Compressor::Zstd(_));
+        // Calculate the sha256 of the archive as we write it into the compressor, so that we don't
+        // need to keep two copies in memory simultaneously.
+        let mut archive = Sha256Writer::new(&mut self.compressor);
 
         let mut ino_index = 1;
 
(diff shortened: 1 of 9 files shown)


dralley:
@DemiMarie I've just discovered that if you create an RPM with a completely incorrect PAYLOADDIGEST ***or*** PAYLOADDIGESTALT and run `rpm -v --checksig test.rpm` then it will still validate OK and will not print the incorrect checksum header in the output.  However if both are wrong, ***then*** it will print out that both are wrong.

Screenshot from 2023-04-15 23-33-44
Screenshot from 2023-04-15 23-34-15
Screenshot from 2023-04-15 23-34-51

Verified that both tags are present in the header

```
rpm -qp --xml test.rpm
.......
.......
  <rpmTag name="Payloaddigest">
	<string>74d1f46600ed240f1ef784f927d8e49366fa6b00a5f13fbb5412275119a719cc</string>
  </rpmTag>
  <rpmTag name="Payloaddigestalgo">
	<integer>8</integer>
  </rpmTag>
  <rpmTag name="Payloaddigestalt">
	<string>4453e31b92551be11ab8aa54f5ad5309eef1f66d66eba2b62ad2de147c96faaa</string>
  </rpmTag>
</rpmHeader>
```

```

That seems wrong.  This is on Fedora 36 but I will check also with Fedora 38

**Comment by dralley** [comment]
It looks like it was intentional but it makes no sense to me https://github.com/rpm-software-management/rpm/pull/880/files#diff-3496c36651db166fa48735c0d33ef48c537c18003e272e3381c16e637fc0d8dcR487-R490

**Comment by dralley** [comment]
Filed `https://github.com/rpm-software-management/rpm/issues/2486`

File: src/rpm/package.rs [hunk 1]
```diff
@@ -294,7 +312,7 @@ impl RPMPackageMetadata {
         Ok(())
     }
 
-    pub fn get_digests(&self) -> Result<Digests, RPMError> {
+    fn get_digests(&self) -> Result<Digests, RPMError> {
```
dralley, line 315: Problem: all these digests are / have to be calculated in different places, and this `Digests` struct is shared with the creation code, so we can't just add them here.  Need to rework this a bit so to reduce churn let's not expose it yet.

src/rpm/headers/types.rs
```diff
@@ -315,6 +316,35 @@ impl Dependency {
     }
 }
 
+pub struct Sha256Writer<W> {
```
drahnr, line 319: Missing docs :)

src/rpm/package.rs
```diff
@@ -120,15 +112,13 @@ impl RPMPackage {
         HACC: std::io::Read,
     {
         let digest_md5 = {
-            use md5::Digest;
```
drahnr, line 123: :+1:

src/rpm/package.rs
```diff
@@ -216,31 +208,59 @@ impl RPMPackage {
         );
 
         verifier.verify(header_bytes.as_slice(), signature_header_only)?;
+        self.verify_digests()?;
```
drahnr, line 211: Is this intended to always verify digests? If they're missing, the verification would fail and the signatures themselves would already contain the content.

I'd be in favor of providing a on-stop-shop `fn verify` that verifies whatever is in the package.

src/rpm/package.rs
```diff
@@ -216,31 +208,59 @@ impl RPMPackage {
...
+            .metadata
+            .header
+            .get_entry_data_as_u32(IndexTag::RPMTAG_PAYLOADDIGESTALGO);
+
+        if let (Ok(payload_digest_val), Ok(payload_digest_algo)) =
+            (payload_digest_val, payload_digest_algo)
+        {
+            let mut hasher = match payload_digest_algo {
+                a if a == FileDigestAlgorithm::Sha2_256 as u32 => sha2::Sha256::default(),
+                _ => unimplemented!("An unknown payload digest algorithm was used"),
```
drahnr, line 247: Should be a regular error, not a panic

src/rpm/package.rs
```diff
@@ -216,31 +208,59 @@ impl RPMPackage {
...
+                // At the present moment even rpmbuild only supports sha256
+            };
+            let payload_digest = {
+                hasher.update(self.content.as_slice());
+                hex::encode(hasher.finalize())
```
drahnr, line 252: Could we document the format this is expected in. If I remember correctly, this is now the third variant.

src/rpm/package.rs
```diff
@@ -216,31 +208,59 @@ impl RPMPackage {
...
+            if payload_digest != payload_digest_val[0] {
+                return Err(RPMError::DigestMismatchError);
+            }
         }
+
+        if recreated_from_content != sig_header_digests {
+            return Err(RPMError::DigestMismatchError);
+        }
+
+        Ok(())
```
drahnr, line 263: With the lack of a response type, we don't know if an actual verification took place or if the items were missing. Depending on the context either might be acceptable or not.

drahnr (changes requested):
A few nits and details to be clarified, other than that, lgtm

src/rpm/package.rs
dralley: Added an @todo item

src/rpm/package.rs
dralley: Yup, I'm just stating that as a practical matter it's hard to share this in both places.  I could make the creation code return a tuple instead and free this up to be user-facing?

src/rpm/package.rs
drahnr: I see, but we do verify both, at least that's what I remember.

src/rpm/package.rs
```diff
@@ -216,31 +208,94 @@ impl RPMPackage {
...
 
         verifier.verify(header_and_content_cursor, signature_header_and_content)?;
 
-        Ok(SignatureVerificationOutcome::Pass)
+        Ok(())
     }
 
-    pub fn verify_digest(&self) -> Result<DigestVerificationOutcome, RPMError> {
+    /// Verify any digests which may be present in the RPM headers
+    pub fn verify_digests(&self) -> Result<(), RPMError> {
```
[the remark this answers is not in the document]
dralley: Or maybe just return `DigestReport` where
```Rust
struct DigestReport {
     payloaddigest: DigestStatus,
     sha256header: DigestStatus,
     sha1header: DigestStatus,
     // etc.
}

enum DigestStatus {
    Success,
    NotPresent,
    Mismatch(String, String),
}
```

with an implementation of `From<DigestStatus>` for `RPMError` which just goes through the list and checks them.

src/rpm/package.rs
dralley: I don't think so, we verify old signatures and new signatures, but that's on a different dimension from the actual type of signatures.  There's RSAHEADER and DSAHEADER, but we don't have any code that reads DSAHEADER

src/rpm/package.rs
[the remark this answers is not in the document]
drahnr: I prefer the latter :+1:

src/rpm/package.rs
```diff
@@ -216,31 +208,94 @@ impl RPMPackage {
...
+        }
+
+        let payload_digest_val = self
+            .metadata
+            .header
+            .get_entry_data_as_string_array(IndexTag::RPMTAG_PAYLOADDIGEST);
+        let payload_digest_algo = self
+            .metadata
+            .header
+            .get_entry_data_as_u32(IndexTag::RPMTAG_PAYLOADDIGESTALGO);
```
drahnr, line 275:
```suggestion
        let payload_digest_algo = FileDigestAlgorithm::try_from(self
            .metadata
            .header
            .get_entry_data_as_u32(IndexTag::RPMTAG_PAYLOADDIGESTALGO))?;
```

src/rpm/package.rs
```diff
@@ -216,31 +208,94 @@ impl RPMPackage {
...
+        {
+            let mut hasher = match payload_digest_algo {
+                a if a == FileDigestAlgorithm::Sha2_256 as u32 => sha2::Sha256::default(),
+                a => {
+                    return Err(RPMError::UnsupportedFileDigestAlgorithm(
+                        FileDigestAlgorithm::from_u32(a)
+                            .unwrap_or_else(|| panic!("unexpected digest algorithm id {}", a)),
+                    ));
+                } // At the present moment even rpmbuild only supports sha256
+            };
```
drahnr, line 288:
```suggestion
            let mut hasher = match payload_digest_algo {
                FileDigestAlgorithm::Sha2_256 => sha2::Sha256::default(),
                 // At the present moment even rpmbuild only supports sha256
            };
```

drahnr: And move the error handling into the `TryFrom for FileDigestAlgorithm` implementation.

src/rpm/builder.rs
```diff
@@ -675,7 +682,7 @@ impl RPMBuilder {
             "4.0-1".to_string(),
         ));
 
-        if matches!(self.compressor, Compressor::Zstd(_)) {
```
drahnr, line 678: If we impl `fn as_dependency(&self) -> Option<Dependency>` for the compressor we can simply call `self.requires.extend(compressor.as_dependency())` which is much cleaner.

drahnr (changes requested):
A few idiomatic adjustments, other than that, :+1:

src/rpm/builder.rs
dralley: I'm not sure that pattern would generalize well enough to be worthwhile.  In fact this might be the only case where that might be possible.

dralley:
@drahnr Completely unrelated question.  I have an implementation of RPM version comparison which I would like to upload.  I could add it to this project, or we could make a new crate for it.  I kind of prefer he latter.

1) Do you think that makes sense?
2) If so, can I have permissions to create new repos in the org or transfer repos into the org?

drahnr:
If possible, make it a separate crate, add it to this repository, add a `workspace` entry to the `Cargo.toml`. That's imho the least churn/value.

drahnr (approved):

PR #131, opened 2023-04-24
Functionality for comparing EVRs / versions

dralley:
### 📜 Checklist


src/version_compare.rs
```diff
@@ -0,0 +1,443 @@
...
+        &self.release
+    }
+
+    pub fn values(&self) -> (&str, &str, &str) {
+        (&self.epoch, &self.version, &self.release)
+    }
+
+    pub fn parse_values(evr: &'a str) ->(&'a str, &'a str, &'a str) {
+        let (epoch, vr) = evr.split_once(':').unwrap_or(evr.split_at(0));
+        let (version, release) = vr.split_once('-').unwrap_or((vr, ""));
```
dralley, line 41: Two ways to do this:

* No real validation on the inputs.  Almost never are you creating these from random values, they're coming from packages you already have.    `librpm` does this.  Nothing about the EVR comparison algorithm actually cares if there's a release or if the version is just a `.` so any restrictions we place on top of that would be basically artificial.
* We put some restrictions in place, like that `epoch` has to be only numeric values, release can't be empty, etc.

dralley:
@drahnr So, you're partial to subcrates over creating a new repo?

I would also like to maybe write a few simple utilities for e.g.

* dumping the tags present & values of RPM headers
* doing low-level lints on RPM headers
* re-packing RPMs

I assume you feel similarly about such a set of `rpm-utils`?

drahnr:
I think all of them could become part of this repository in a single workspace, but ultimately I'll leave that up to you

dralley:
I'm not *against* making this its own crate, but I'm not really in favor of doing so immediately.  Maybe a bit further down the line.

src/version.rs
```diff
@@ -0,0 +1,934 @@
...
+/// Release indicates firstly the number of times this package has been released - for instance,
+/// with custom patches and backports not present in the upstream, but may also indicate other
+/// details such as the OS it was built for (fc38, el9) or portions of a git commit hash.
+///
+/// Architecture indicates the CPU architecture that this package is intended to support.
+///
+/// In many contexts (on a system, in a repository), package NEVRAs are meant to be unique. You can have
+/// different packages with the same NEVRA - but you can't install both, or put them both in a repo.
+#[derive(Clone, Debug, Default, Eq, PartialEq)]
+pub struct NEVRA<'a> {
```
drahnr, line 25: Can we avoid all caps types?

src/version.rs
```diff
@@ -0,0 +1,934 @@
...
+            self.name,
+            self.evr.as_normalized_form(),
+            self.arch
+        )
+    }
+}
+
+impl<'a> fmt::Display for NEVRA<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}-{}.{}", self.name, self.evr, self.arch)
```
drahnr, line 122: Is there a particular reason we are not using the normalized form, always?

drahnr (changes requested):
Two nits, otherwise LGTM

src/version.rs
dralley: The short form is actually more common, it's only in specific places like repo metadata where the normalized form ends up being used.

src/version.rs
dralley: NEVRA is an acronym, is the main reason.  I can change it, but then for instance there will be a difference between the name as used in the docs (where all caps is standard for such acronyms) and the type name.

drahnr: That's fine, acronyms are either to be camel (so `Nevra`) cased or named to something that can be interpreted, i.e. `PackageVersionFull` or similar.

src/version.rs
dralley: The current implementation works the same way that `rpm -qp --queryformat="%{NEVRA}" $pkg` does

PR #133, opened 2023-04-30
Add SHA256HEADER tag to signature header for built packages (and fix the verification code)

dralley:
### 📜 Checklist

### DIFF
--- a/src/rpm/headers/header.rs
+++ b/src/rpm/headers/header.rs
@@ -341,7 +341,7 @@ impl Header<IndexSignatureTag> {
     ) -> Result<Header<IndexSignatureTag>, RPMError> {
         let result = Self::parse_async(input).await?;
 
-        let padding = 8 - (result.index_header.data_section_size % 8);
+        let padding = (8 - (result.index_header.data_section_size % 8)) % 8;
         if padding > 0 {
             let mut discard = vec![0; padding as usize];
             input.read_exact(&mut discard).await?;
@@ -355,7 +355,7 @@ impl Header<IndexSignatureTag> {
         let result = Self::parse(input)?;
         // this structure is aligned to 8 bytes - rest is filled up with zeros.
         // if the size of our store is not a modulo of 8, we discard the padding bytes
-        let padding = 8 - (result.index_header.data_section_size % 8);
+        let padding = (8 - (result.index_header.data_section_size % 8)) % 8;
         if padding > 0 {
             let mut discard = vec![0; padding as usize];
             input.read_exact(&mut discard)?;
@@ -370,7 +370,7 @@ impl Header<IndexSignatureTag> {
     ) -> Result<(), RPMError> {
         self.write_async(out).await?;
         // align to 8 bytes
-        let padding_needed = 8 - (self.index_header.data_section_size % 8);
+        let padding_needed = (8 - (self.index_header.data_section_size % 8)) % 8;
         if padding_needed > 0 {
             let padding = vec![0; padding_needed as usize];
             out.write_all(&padding).await?;
@@ -381,7 +381,7 @@ impl Header<IndexSignatureTag> {
     pub(crate) fn write_signature<W: std::io::Write>(&self, out: &mut W) -> Result<(), RPMError> {
         self.write(out)?;
         // align to 8 bytes
-        let padding_needed = 8 - (self.index_header.data_section_size % 8);
+        let padding_needed = (8 - (self.index_header.data_section_size % 8)) % 8;
         if padding_needed > 0 {
             let padding = vec![0; padding_needed as usize];
             out.write_all(&padding)?;
--- a/src/rpm/headers/header.rs
+++ b/src/rpm/headers/header.rs
@@ -321,11 +321,12 @@ impl Header<IndexSignatureTag> {
         headers_plus_payload_size: usize,
         md5sum: &[u8],
         sha1: &str,
+        sha256: &str,
         rsa_spanning_header: &[u8],
         rsa_spanning_header_and_archive: &[u8],
     ) -> Self {
         SignatureHeaderBuilder::new()
-            .add_digest(sha1, md5sum)
+            .add_digest(sha1, sha256, md5sum)
             .add_signature(rsa_spanning_header, rsa_spanning_header_and_archive)
             .build(headers_plus_payload_size)
     }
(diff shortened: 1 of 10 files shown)


src/rpm/package.rs
```diff
@@ -84,49 +84,28 @@ impl RPMPackage {
...
-
-    // @todo: delete when EL7 goes EOL, PAYLOADDIGEST has been available for years now.
+    // @todo: delete most of this when EL7 goes EOL, the combination of SHA256HEADER and PAYLOADDIGEST
+    // have been available for a while.  EL7 has SHA1HEADER but not PAYLOADHEADER so the only digest which
+    // can verify both is SIGMD5. EL8+ (and similar distros) have both, so SHA1HEADER isn't needed anymore.
     /// Prepare both header and content digests as used by the `SignatureIndex`.
-    pub(crate) fn create_legacy_header_digests<HC, HACC>(
-        header_cursor: HC,
-        header_and_content_cursor: HACC,
-    ) -> Result<Digests, RPMError>
```
dralley, line 109: This was inefficient because it forced copying everything into a new buffer and repeatedly calling `read()`.  But we already have it all in memory.

src/rpm/package.rs
```diff
@@ -240,24 +216,19 @@ impl RPMPackage {
...
-            if sha1 != recreated_from_content.header_digest {
+            if sha1 != pkg_actual_digests.header_digest_sha1 {
                 return Err(RPMError::DigestMismatchError);
             }
         }
 
         if let Ok(sha256) = sha256 {
-            let sha256_calculated = {
-                let mut hasher = sha2::Sha256::default();
-                hasher.update(self.content.as_slice());
```
dralley, line 257: I was verifying the wrong thing here.  Existing tests don't catch it because EL7 predates SHA256HEADER

dralley:
Something is wrong with the signature header padding code such that adding a new entry there, or even removing an existing one, causes `rpm` to mispredict where the header is supposed to  be (`hdr magic: BAD`)

src/rpm/package.rs
drahnr: The function was meant to precisely avoid the need to have anything in memory and work on a reasonable chunk size in memory without having to load everything into memory upfront.

src/rpm/package.rs
```diff
@@ -84,49 +86,28 @@ impl RPMPackage {
...
+    // can verify both is SIGMD5. EL8+ (and similar distros) have both, so SHA1HEADER isn't needed anymore.
     /// Prepare both header and content digests as used by the `SignatureIndex`.
-    pub(crate) fn create_legacy_header_digests<HC, HACC>(
-        header_cursor: HC,
-        header_and_content_cursor: HACC,
-    ) -> Result<Digests, RPMError>
-    where
-        HC: std::io::Read,
-        HACC: std::io::Read,
-    {
```
drahnr, line 113: Same here. The idea is to migrate away from the in memory requirement, ultimately to somthing akin to `Seek + Read` as minimal bounds.

src/rpm/package.rs
dralley: Until we actually don't keep the `content` in memory, I'd rather we just kept things simple.  By the time we get around to not doing so, we might not need this code at all anymore, that's my point.

src/rpm/headers/header.rs
```diff
@@ -987,6 +989,11 @@ mod test {
                     offset,
                     IndexData::StringTag(sha1.to_owned()),
                 ),
+                IndexEntry::new(
+                    IndexSignatureTag::RPMSIGTAG_SHA256,
+                    offset,
+                    IndexData::StringTag(sha256.to_owned()),
```
dralley, line 995: If the length of this string is 0, 8, 12, 16, 24, 32, 40, 48 (I stopped there but you get the gist), RPM fails to parse

dralley: It looks like this carries through other tags.  If you start tweaking the length of the `RSA` signature, it exhibits the same behavior.

With both the SHA1 and SHA2 digests in the header, an RSA tag with length 6 (binary) fails to parse while other lengths parse successfully but fail to validate.

dralley:
Issue resolved

dralley:
MSRV seems problematic, without https://github.com/rust-lang/cargo/issues/9930

dralley:
I have an idea that I think could make both of us happy

dralley:
Actually two ideas.  Before I actually spend a bunch of time on it, here they are:

1) We could do something similar to what `rpm` does, and provide a wrapper for the reader which is aware of what digests need to be generated, and generates them automatically as the file is read.  You would just need to provide a little bit of information about what portion of the package it is currently processing, i.e. "currently processing the header" or "currently processing the content".

2) We could implement header-only package reading - that is, read a package without storing the content in memory.  And then, require `Seek` as you described, and store the segment boundaries, and produce the digests as-needed from the on-disk data.

Either approach I would consider superior to what we currently do, because we parse the headers and then rewrite them and take that digest, rather than taking the digest of the data which is on-disk.  That creates some additional surface for bugs.

I would kind of prefer that this doesn't block the release though.  Are you satisfied with having a plan to move forwards?

dralley:
@drahnr ^

drahnr:
I agree that the re-reading is not necessary, and that hash-while-read is the preferred approach. In my mind, I don't see any of this blocking a release.

src/rpm/headers/header.rs
```diff
@@ -365,7 +373,7 @@ impl Header<IndexSignatureTag> {
     ) -> Result<(), RPMError> {
         self.write_async(out).await?;
         // align to 8 bytes
-        let padding_needed = 8 - (self.index_header.data_section_size % 8);
+        let padding_needed = (8 - (self.index_header.data_section_size % 8)) % 8;
```
drahnr, line 376: nit: Maybe worth a tiny `const fn padding_needed(len: usize) -> usize {..}`

src/rpm/package.rs
drahnr: I'd feel strongly about it, it's the next think I want to tackle, given a find a few hours to work on it

drahnr (approved):
Two nits, it'd be nice if the `Reader` based `fn`s could stay so I don't have to partial revert again in the next weeks once I get around avoiding to read it all at once.

src/rpm/headers/header.rs
dralley: `data_section_size` isn't known at compile time so I don't think it would ever be executed in `const` context

dralley:
I was doing some partial prototyping of it and I don't think that with either of the proposed solutions actually benefits from having the reader-based code in place.

PR #136, opened 2023-05-12
Allow users to configure compression levels

dralley:
### 📜 Checklist


dralley:
@ikrivosheev

dralley:
About the default compression type: I'm open to changing it, but I really do think we need to opt into some kind of compression by default.  I'm in a position to see some of the infrastructure hosting costs of large repos, and have spoken with users that manage terabytes of RPMs, so to me the cost/benefit ratio of the various defaults is very stark in favor of some kind of compression.

Zstd would be ideal but so long as CentOS 7 is our minimum bound we cannot choose it.  So until then, Gzip is probably the best tradeoff.

src/rpm/builder.rs
```diff
@@ -168,7 +168,7 @@ impl RPMBuilder {
     ///
     /// ```
     /// # fn foo() -> Result<(), Box<dyn std::error::Error>> {
-    /// let pkg = rpm::RPMBuilder::new("foo", "1.0.0", "MPL", "x86_64", "some bar package")
+    /// let pkg = rpm::RPMBuilder::new("foo", "1.0.0", "MIT", "x86_64", "some bar package")
```
drahnr, line 171: That wasn't a typo :)

src/rpm/builder.rs
```diff
@@ -198,28 +198,62 @@ impl RPMBuilder {
...
+    ///
+    /// The a changelog entry consists of an entry name (which includes author, email followed by
+    /// a dash followed by a version number), description, and the date and time of the change.
 
     /// ```
     /// # fn foo() -> Result<(), Box<dyn std::error::Error>> {
     /// use rpm::chrono::TimeZone;
     ///
-    /// let pkg = rpm::RPMBuilder::new("foo", "1.0.0", "Apache-2.0", "x86_64", "some baz package")
+    /// let pkg = rpm::RPMBuilder::new("foo", "1.0.0", "MIT", "x86_64", "some baz package")
```
drahnr, line 256: I don't understand why all examples are unified to MIT?

drahnr (changes requested):
Besides the two odd stray changes, LGTM

src/rpm/builder.rs
dralley: It's just shorter

src/rpm/builder.rs
dralley: It's not a correct SPDX identifier though, I could change it to `MPL-2.0`

drahnr: Yes please

src/rpm/builder.rs
drahnr: Please revert

dralley:
Wdyt about gzip vs. zstd as the default?  Knowing that the latter can't be used on EL7

drahnr:
I am still leaning towards `None`, but I don't have a strong opinion. It'd be nice to at least not break anything that's currently officially supported/maintained by default to avoid any surprises.

src/rpm/compressor.rs
```diff
@@ -73,32 +75,45 @@ impl Compressor {
...
-            Compressor::Gzip(encoder) => Ok(encoder.finish().into_result()?),
+            Compressor::Gzip(encoder) => Ok(encoder.finish()?),
             Compressor::Zstd(encoder) => Ok(encoder.finish()?),
             Compressor::Xz(encoder) => Ok(encoder.finish()?),
         }
     }
+}
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum CompressionDetails {
```
dralley, line 86: Should this be named `CompressionWithLevel`

tests/compat.rs
```diff
@@ -222,7 +222,7 @@ mod pgp {
                 "x86_64",
                 "spins round and round",
             )
-            .compression(CompressionType::Gzip)
+            .compression(CompressionType::Zstd)
```
dralley, line 225: Just trying to increase coverage on the compat tests, we use Gzip already on one of them.

ikrivosheev (commented):
@dralley hello! It looks good, but I don't see an API for set compression level in `Compressor`.

dralley:
@ikrivosheev  See the 2nd example provided in the documentation

ikrivosheev:
@dralley for me it looks strange. We have two public types for one operation. Users need to think about it.

dralley:
Imo it's not so different from APIs which take AsRef<str> or Into<Cow<str>>

dralley:
>I think is big different between AsRef and this API. When you use AsRef this mean that you want to work with objects like for example: Vec<u8>, [u8], Mmap, ... These objects are not similar but they have a similar interface.    Here you have a very similar objects.

I don't see that as a big difference.  Besides, a `&str` and `String` and a `Path` and a `PathBuf` are very similar objects and they will all work in an `AsRef<Path>` parameter

ikrivosheev:
> > I think is big different between AsRef and this API. When you use AsRef this mean that you want to work with objects like for example: Vec, [u8], Mmap, ... These objects are not similar but they have a similar interface.    Here you have a very similar objects.
> 
> I don't see that as a big difference. Besides, a `&str` and `String` and a `Path` and a `PathBuf` are very similar objects and they will all work in an `AsRef<Path>` parameter

We need another opinion)

drahnr (approved):
I agree that this is not the must idiomatic interface, generally the decompressor doesn't care about it much, for the compressing side only the actual compression `fn` _must_ know about it. Now the need to provide the compression level as header causes the whole dependency and I don't think there is _great_ solution for it given that requirement.

The impl is a _good_ solution :+1: 

---

`AsRef<Path>` liberates the caller from doing extra conversions, so I am all for that.

PR #137, opened 2023-05-13
Make parsing more robust in the face of unknown or unexpected tags

### LINKED ISSUE #119: Raises an error during parsing when encountering unknown tags
If an RPM being parsed contains any tag which isn't in the hardcoded enum definition for `IndexTag` or `IndexSignatureTag`, an error is raised.

That's a bit problematic from a forwards compatibility perspective, there's no real reason to raise an error in that situation because we could otherwise roundtrip just fine.  The mere fact that an unknown tag exists shouldn't generally prevent it from being read or written.

The challenge will be making that happen without losing the nice ergonomics and debug output of the enum.

One possible approach: have e.g. an (possibly #[repr(u32)]) enum ID { ... } that only contains currently known values and a #[repr(transparent)] struct RawID { value: u32 } that supports all values,and have RawID: From<ID> and ID: TryFrom<RawID, Error = UnrecognizedID>

dralley:
Unfortunately it is common for certain tags to be present in the wrong header, so there are many packages in the wild which rpm-rs would fail to parse. Additionally any packages built with brand-new versions of rpmbuild which contain tags not listed in our list of constants would fail.

In either case, we ought to be a bit more tolerant. Users should be able to parse the RPMs and do basic things with them even if we don't provide the ability to work with the unknown parts of the header.

This PR turns the `tag` into a standard u32 rather than an enum and removes the `InvalidTag` error variant.  It retains the other properties of `Header<T>` using the `PhantomData` zero-sized-type.  Additionally it adjusts the `Debug` trait impl to continue showing the tag names as before, rather than the raw integer value.

closes #119

### 📜 Checklist

### DIFF
--- a/src/rpm/headers/header.rs
+++ b/src/rpm/headers/header.rs
@@ -657,7 +657,6 @@ pub(crate) struct IndexEntry<T: num::FromPrimitive> {
     pub(crate) data: IndexData,
     pub(crate) offset: i32,
     pub(crate) num_items: u32,
-
     // Marks what type of IndexEntry it is
     entry_type: PhantomData<T>,
 }
@@ -665,15 +664,15 @@ pub(crate) struct IndexEntry<T: num::FromPrimitive> {
 /// Custom Debug impl for the benefit of showing the tag name, if we are familiar with it
 impl<T: Tag> std::fmt::Debug for IndexEntry<T> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let tag = T::from_u32(self.tag);
-        let tag_val = if let Some(val) = tag {
+        let known_tag: Option<T> = num::FromPrimitive::from_u32(self.tag);
+        let tag_name = if let Some(val) = known_tag {
             format!("{:?}", val)
         } else {
             format!("UnknownTag[{:?}]", self.tag)
         };
 
         f.debug_struct(&format!("IndexEntry<{}>", T::tag_type_name()))
-            .field("tag", &tag_val)
+            .field("tag", &tag_name)
             .field("data", &self.data)
             .field("offset", &self.offset)
             .field("num_items", &self.num_items)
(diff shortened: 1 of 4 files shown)


src/rpm/headers/header.rs
```diff
@@ -311,6 +312,38 @@ where
     }
 }
 
+impl fmt::Display for Header<IndexSignatureTag> {
```
dralley, line 315: Makes it easy to dump a readable summary of the package (the Debug impl isn't very readable due to verbosity)

```
Signature Header - Entries: 11,  Data Section Size: 4212 bytes
===================================================
    HEADER_SIGNATURES: Bin
    RPMSIGTAG_RSA: Bin
    RPMSIGTAG_SHA1: String
    RPMSIGTAG_SHA256: String
    RPMSIGTAG_SIZE: i32
    RPMSIGTAG_PGP: Bin
    RPMSIGTAG_MD5: Bin
    RPMSIGTAG_PAYLOADSIZE: i32
    RPMTAG_INSTALLTIME: Bin
    << UnknownTag >> [5090]: StringArray
    << UnknownTag >> [5091]: i32

Header - Entries: 62,  Data Section Size: 5177 bytes
===================================================
    RPMTAG_HEADERIMMUTABLE: Bin
    RPMTAG_HEADERI18NTABLE: StringArray
    RPMTAG_NAME: String
    RPMTAG_VERSION: String
    RPMTAG_RELEASE: String
    RPMTAG_SUMMARY: I18NString
    RPMTAG_DESCRIPTION: I18NString
    RPMTAG_BUILDTIME: i32
    RPMTAG_BUILDHOST: String
    RPMTAG_SIZE: i32
   ...
   ...
```

src/rpm/headers/header.rs
```diff
@@ -654,6 +709,8 @@ impl<T: Tag> IndexEntry<T> {
                 data,
                 offset,
                 num_items,
+
```
cmeister2, line 712: nit: empty line

src/rpm/headers/header.rs
```diff
@@ -612,34 +651,50 @@ impl IndexHeader {
...
+    // Marks what type of IndexEntry it is
+    entry_type: PhantomData<T>,
+}
+
+/// Custom Debug impl for the benefit of showing the tag name, if we are familiar with it
+impl<T: Tag> std::fmt::Debug for IndexEntry<T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let tag = T::from_u32(self.tag);
+        let tag_val = if let Some(val) = tag {
```
cmeister2, line 669: This code feels a bit magical to me, maybe I'm just a bit removed from it. I _think_ because `IndexTag` derives FromPrimitive that the `from_u32` code will actually give us something useful. Not a problem cause it compiles, it's just more opaque than I expected for Rust. I had to load it into an editor to chase down the traits.

cmeister2 (commented):
Few minor comments from me.

src/rpm/headers/header.rs
dralley: That's correct.  Want me to call the trait explicitly as was done here? https://github.com/rpm-rs/rpm/pull/137/files#diff-e8b11ac65b082c9412d618eb40c89f86b44fada22b3dbc251ad1d289d72770fbL629

dralley:
@cmeister2 Updated

Ignore the other commits, this is currently rebased on top of the other PR which needs to be merged first

cmeister2:
Only thing that would make this better is a test showing parsing being better in the face of an unknown tag...

PR #139, opened 2023-05-15
Replaced `SeqCursor` to chain of `io::Cursor`

ikrivosheev:
### 📜 Checklist

Hello! I replaced `SeqCursor` to chain of `io::Cursor`


CHANGELOG.md
```diff
@@ -47,6 +47,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   Users can configure it with `RPMBuilder::build_time()`.
 - Improved support for packages >4gb
 - Always write tags in sorted order
+- Changed signature for `RPMPackage::parse`. Now it apply `io::Read` instead `io::BufRead`
```
dralley, line 50: I'm a little conflicted on this, looser constraints are usually nice but buffered IO is something you generally want to encourage to avoid your code being much slower than expected.  And we're doing a bunch of small reads to parse the headers.  With BufReader, the whole package is 99% likely to fit into the buffer (8mb default) so it's all coalesced into one actual read syscall.

ikrivosheev: @dralley I totally agree! But, This does not mean that library cannot work with `io::Read`. If we can give options for work with simple `io::Read` than why not) U
ser can make a decision use `io::Read` or `io::BufRead`. For example, I use my own impl of working with file (If file is smaller then 5mb then I load it into memory) and then call `RPMPackage::parse`. I don't need to use `io::BufRead` but I have to wrap this into `io::BufRead` for working with RPMPackage.

ikrivosheev: It's ok. If programmer care about performance  than he (or she) cans use `io::BufReader` and everything continue working)  It's not a breaking change)

ikrivosheev: https://github.com/rpm-rs/rpm/blob/f6e62aa3c2f27130e44beff8f0c91a3bdcef7bc5/src/rpm/package.rs#L55 here we are working with `io::BufReader`. This PR is about convenience, not about performance)

dralley: > For example, I use my own impl of working with file (If file is smaller then 5mb then I load it into memory) 

That doesn't seem much different from just using a `BufReader` with a size of 5mb?  Are you trying to keep the memory really low or something?

I'm fine with all of that but my misgiving is that if we later find a use case where it would have been nice to use `BufRead`, which we might depending on how we implement the whole not-loading-the-payload-immediately thing, then we'd need to change the signature back.

ikrivosheev: I cannot use `BufRead`. My code looks like:

```rust
enum Storage {
    Memory(Vec<u8>),
    Mmap(Mmap),
}

pub struct CachedFile(Cursor<Storage>);

impl AsRef<[u8]> for CachedFile {...}
impl Read for CachedFile {...}
impl Seek for CachedFile {...}
```

I need to support both interfaces: buffer and reader.

ikrivosheev:
> Please use rebases instead of merges fwiw.

Done

dralley:
I still have mixed feelings but I'm not strictly against it.  Will wait for an opinion from @cmeister2 or @drahnr 

I am curious whether you actually see benefit from using mmap, my understanding is that it's actually quite situational and loses out to normal reads if you aren't playing to its benefits.

ikrivosheev:
> I still have mixed feelings but I'm not strictly against it. Will wait for an opinion from @cmeister2 or @drahnr
> 
> I am curious whether you actually see benefit from using mmap, my understanding is that it's actually quite situational and loses out to normal reads if you aren't mapping the same file into multiple threads / processes, or reading only a small portion of it, or skipping back and forth to random locations.

I'm working on antivirus. I do a lot stuff with file, for example: 
1. Rule engine (for binary signatures) needs buffer like object
2. Unpackers need `Read` like interface (some buffer like object)

As I know, mmap preloads about 14 pages.

cmeister2:
FWIW for my use case I ended up having to do:
```
    /// Attempts to create a `RpmRemotePackage` from something that impls
    /// Read.
    pub fn new_from_read<R: Read>(reader: R) -> Result<Self, PkgError> {
        let mut bufread = BufReader::new(reader);
        let metadata = RPMPackageMetadata::parse(&mut bufread)?;
        Ok(Self { metadata })
    }
```
for the code on `master`. I think generally speaking you're more likely to have `impl Read` than `impl BufRead`.

(As an aside I somehow didn't need a mutable Reader with my previous code - I haven't investigated why this is, yet)

cmeister2:
> (As an aside I somehow didn't need a mutable Reader with my previous code - I haven't investigated why this is, yet)

This is because I wrapped the Reader in a BufReader myself! https://docs.rs/fez/0.2.0/src/fez/rpm/package.rs.html#251

dralley:
I always just do the wrapping inline when necessary, e.g. `function(&mut BufReader(reader))`

It doesn't really take more space than a special function.

CHANGELOG.md
cmeister2: Just to correct something above - the internal buffer for BufReader is 8KB, not 8MB. 8KB is less likely to hold an entire RPM package.

cmeister2:
Reading https://www.brandons.me/blog/bufread-and-when-to-use-it, and the general upshot is "it's an optimization, benchmark it to be sure". However, they also mention that if you're repeatedly reading a large number of bytes (i.e. loading the payload) then BufReader won't help as much (and is likely to just be an extra unnecessary memory copy).

I think the question boils down to "how do we use the traits here", and I think we might actually get more mileage out of `Read`, simply because we're generally just doing `read_exact()` and `read_to_end()`, so there's no need for the wrapper of BufRead.

That's my reading of it, anyway!

dralley:
Well, I suppose the answer is benchmarking.  But I would expect that on average it's slower due to all of the tiny header reads.  It would be possible to end up underwater if the payload is large enough but usually packages are in the kilobytes range so I doubt it's the case.

My main concern isn't performance but whether we'll end up needing to implement buffering separately anyway for certain operations.

cmeister2:
> Did you have other Fez patches that you think would be useful to upstream?

The main difference between rpm-rs and fez 0.2.0 was the "on demand" loading class - but that's been made redundant by the exposing of RPMPackageMetadata, which is a much more elegant solution to the problem.

drahnr:
I think the entire topic doesn't warrant the discussion it takes. Most internal `fn`s still create `BufReader` wrapping, the only place where it _could_ matter is public API. I am in favor of enforcing the `BufRead` trait bound, just for the sake of preventing possible footguns. The additional pain is low regardless. If `Read` is the bound, the documentation should recommend to use a `BufRead` unless disproven by benchmarks.

ikrivosheev:
> Ack. @ikrivosheev does your use case 100% require this? If not, could you present some data on situations where it would be beneficial to use `Read` directly?
> 
> We can go ahead and merge these cleanups if you want to split that change off into it's own smaller commit for resubmission later, if it ends up making sense.

@dralley well... I removed changes with `io::Read`.

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Removed
```
dralley, line 9: This wasn't public API to begin with so I don't think it needs a changelog personally.

dralley: Otherwise this LGTM

ikrivosheev: Removed

drahnr (approved):
I must have missed the existence of `fn chain(..)` when impl seq cursor.

PR #141, opened 2023-05-18
0.11.0 release

dralley:


src/rpm/package.rs
```diff
@@ -626,10 +626,18 @@ impl RPMPackageMetadata {
 
     #[inline]
     pub fn get_payload_compressor(&self) -> Result<CompressionType, RPMError> {
```
dralley, line 628: Really inconvenient to get an error instead of `CompressionType::None` on uncompressed packages.

drahnr: It depends how the absence of the value is interpreted. It'd be fine if the error was only used in case of unknown vales rather than absence which (afaiu) is perfectly valid.

dralley: If you're ok with it then :+1:

Can you approve?

cmeister2: I'm surprised this is a bug fix inside of the 0.11.0 branch release; normally I'd expect a bug fix to be in its own PR and branch.

dralley: I know, it's not the best thing.  It's just a small thing I noticed that I wanted to fix before cutting the release

dralley: Ok... one more change :P

src/rpm/headers/types.rs
```diff
@@ -15,17 +15,17 @@ pub struct RPMPackageSegmentOffsets {
...
+    pub sha_checksum: String,
+    pub link: String,
+    pub flags: FileFlags,
+    pub user: String,
+    pub group: String,
+    pub base_name: String,
     #[allow(unused)]
-    pub(crate) dir: String,
+    pub dir: String,
     pub(crate) content: Vec<u8>,
```
dralley, line 29: `get_file_entries()` is pretty pointless if none of the fields are accessible (and there are no accessors)

src/rpm/package.rs
drahnr: yes please

src/rpm/headers/types.rs
drahnr: Could you remove the allow unused annotation too?

dralley: done

dralley:
@cmeister2 LGTY?

cmeister2 (approved):
LGTM

PR #142, opened 2023-05-21
Make fields on FileEntry and other structs public

dralley:
### 📜 Checklist


cmeister2:
LGTM - what's the motivation behind the change?

dralley:
>LGTM - what's the motivation behind the change?

@cmeister2 Could you approve?  The motivation is that we expose the structs themselves to the user and have functions for getting the data, but they return these structs, and none of the fields on the struct are public, which makes them kind of useless.

Anyway, specifically, I'm writing an analogue of createrepo_c which requires some of this information.

cmeister2 (approved):

PR #146, opened 2023-05-26
Support EdDSA / ed25519 signing and verification

dralley:
### 📜 Checklist


src/rpm/signature/traits.rs
```diff
@@ -8,63 +8,58 @@ use crate::errors::*;
...
 /// Signing trait to be implement for RPM signing.
-pub trait Signing<A>: Debug
+pub trait Signing: Debug
 where
-    A: algorithm::Algorithm,
     Self::Signature: AsRef<[u8]>,
 {
     type Signature;
     fn sign<R: Read>(&self, data: R) -> Result<Self::Signature, RPMError>;
+    fn algorithm(&self) -> AlgorithmType;
```
dralley, line 24: So here's my thinking:

If we are just handing the Signer / Verifier raw bytes so that it can do the PGP key parsing, then it makes no sense to use a compile-time generic for the algorithm type.  Because when it parses the key, it can just return an error if the key is of the wrong type.  And since all the algorithms are handled by the same PGP implementation anyway, it would be silly to duplicate the code.

I'm approximately 100% sure we could make this more ergonomic still, but this is as far as I wanted to go in this PR, we can continue cleaning it up later.

dralley:
@TommyLike

src/rpm/package.rs
```diff
@@ -148,34 +156,54 @@ impl RPMPackage {
     #[cfg(feature = "signature-meta")]
     pub fn verify_signature<V>(&self, verifier: V) -> Result<(), RPMError>
     where
```
dralley, line 158: There are a few open questions:

* If you attempt to verify the signatures of a package with no signatures, what error message should it return?  `RPMError::NoSignatureFound` is used for something different, but the signature of `RPMError::VerificationError`  is designed in such a way that it can't easily be used outside of `Verifier` unless we make a few more changes to that API.  New error type, or adjust the existing ones?

* Lets say the package is EdDSA signed and I attempt to verify using a Verifier built from an RSA public key.  It ought to fail (but currently it doesn't).  But again,  `RPMError::VerificationError`  can't be used easily outside of `Verifier`.  I suppose that we could force it to try to verify the signature even though we know it will fail, which would raise the error we expect?

src/rpm/builder.rs
```diff
@@ -475,7 +475,7 @@ impl RPMBuilder {
     #[cfg(feature = "signature-meta")]
     pub fn build_and_sign<S>(self, signer: S) -> Result<RPMPackage, RPMError>
     where
-        S: signature::Signing<signature::algorithm::RSA>,
```
drahnr, line 478: I don't quite understand why we revert a compile time check back to a runtime check. Instead of removing it the `Algorithm` could have become a generic too.

src/rpm/signature/traits.rs
drahnr: Do we really want automatic derivation of the algorithm from the key? That's my key (no pun intended) question here. I'd imagine that a user wants to use a specific key. The compile time guarantees could still be combined.

```
fn sign_with_key_given(self, key: ..) {
let key = key.parse()?;
match key.algorithm() {
Algorithm::RSA => signer.sign::<Algorithm::RSA>(),
// ..
}
```

now this guarantees compiletime consistency and only picks the path according to the convenience route involving algorithm derivation. People that do have the knowledge of the key type statically could maintain correctness by definition.

That's my thinking. Pushing out the runtime checks as far as possible.

src/rpm/builder.rs
dralley: There was no compile-time check because you don't get to avoid the part where the key itself is of the type you want.  With `build_and_sign::<algorithm::RSA>` you still have to make sure the key you pass in is RSA, except that we would then need to return a runtime error when you pass in a key of the wrong type.

That just doesn't feel helpful at all.

src/rpm/package.rs
dralley: @drahnr Any feedback?

TommyLike:
@drahnr thanks, looking forward to this feature!

dralley:
@TommyLike You can feel free to review also

src/rpm/headers/signature_builder.rs
```diff
@@ -111,21 +111,37 @@ impl SignatureHeaderBuilder<Empty> {
...
+
+    pub fn add_eddsa_signature(
+        mut self,
+        sig_header_only: &[u8],
+    ) -> SignatureHeaderBuilder<WithSignature> {
+        let offset = 0i32; // filled externally later on
+        self.entries.push(IndexEntry::new(
+            IndexSignatureTag::RPMSIGTAG_DSA,
+            offset,
+            IndexData::Bin(sig_header_only.to_vec()),
```
TommyLike, line 144: I am not familiar with this, but why we dont' need 'RPMSIGTAG_GPG' here?

dralley: Because EdDSA support was introduced only a few years ago, and any package which is signed as such ought to have payload digests too, which makes checking the legacy v3 style signatures pointless.

The only reason to check v3 style signatures is that EL7 packages aren't guaranteed to have payload digests, so the only way to verify the integrity of the payload is to use the v3 header+payload style ones.

TommyLike: thanks!

src/rpm/builder.rs
drahnr: What I tried to say, if we can avoid runtime `match`es on the key type, and expect the user to provide the correct one, it allows for a more meta code representation approach, which I'd personally prefer over `match` in quite a few places. The compiler can then enforce a few more things and _users_ that know their key type for sure, can have the `match` _outside_ of `rpm-rs` and we can be happy OR possibly take advantage of the type based definition of key types _iff_ they can based on meta programming on their end, too. 

Reasons pro generic based crypto:

* user can use generics to extend type saftey
* compiler help
* less `match` based clutter
* avoids the `build_and_sign_...` variants which all almost do the same
* could separate concerns of doing the actual signing vs the mechanics of creating the tag

drahnr: Imho we should not _parse_ AND _sign_ in the same function, but provide key types that are external to the i.e. `build_and_sign` call / `Signer`.

dralley:
@drahnr Please take another look, I've added tests for the two cases I mentioned.

dralley:
@drahnr Can I merge this?

TommyLike:
> @drahnr Can I merge this?

Support having this. btw, can we verify this signature by rpm command?

dralley:
Yes, it's just a normal RPM signature and `rpm --checksigs` works with it, it's checkable at installation time, etc.

CHANGELOG.md
```diff
@@ -16,15 +17,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Added
 
 - `PackageBuilder::source_date` method for clamping modification time of files,
-  build time of the package, and signature time. This functionality is required for
+  build time of the package, and signature timestamp. This functionality is required for
   reproducible generation of packages.
-- `Package::sign_with_timestamp` method.
+- `Package::sign_with_timestamp` method.s
```
drahnr, line 22:
```suggestion
- `Package::sign_with_timestamp` method.
```

src/rpm/builder.rs
```diff
@@ -516,31 +516,34 @@ impl PackageBuilder {
             header_digest_sha256,
         } = Package::create_sig_header_digests(header.as_slice(), content.as_slice())?;
 
+        let now = Timestamp::now();
+        let t = match source_date {
```
drahnr, line 520:
```suggestion
        let signature_timestamp = match source_date {
```

src/rpm/headers/header.rs
```diff
@@ -906,64 +889,4 @@ mod test {
 
         Ok(())
     }
-
-    #[cfg(feature = "signature-meta")]
-    #[test]
-    fn signature_header_build() {
```
drahnr, line 912: ntms: got moved

src/rpm/signature/pgp.rs
```diff
@@ -54,31 +66,45 @@ impl traits::Signing<traits::algorithm::RSA> for Signer {
...
+        match secret_key.algorithm() {
+            PublicKeyAlgorithm::RSA => Ok(Self {
+                secret_key,
+                algorithm: AlgorithmType::RSA,
+            }),
+            PublicKeyAlgorithm::EdDSA => Ok(Self {
+                secret_key,
+                algorithm: AlgorithmType::EdDSA,
+            }),
+            a => Err(Error::UnsupportedPGPKeyType(a)),
```
drahnr, line 106:
```suggestion
            algorithm => Err(Error::UnsupportedPGPKeyType(algorithm)),
```

drahnr (approved):
A few nits, I am still not particularly excited about the changeset, but I do understand `RPMSIGTAG_PGP` essentially voids any const generic approaches.

src/rpm/headers/header.rs
dralley: ntms?

dralley: (guessing) note to myself?

dralley:
Yeah I'm not against any further improvements and we should be able to simplify this further in a year.

src/rpm/headers/header.rs
drahnr: Yes,

PR #149, opened 2023-05-31
Use long sizes only for big packages

newpavlov:
Closes #148 and fixes several clippy lints.

Issue #148, opened 2023-05-30
Compatability with CentOS 7


newpavlov:
By using the same library and snippet as in #147 I get [`libdummyadd.rpm`](https://github.com/rpm-rs/rpm/files/11604258/libdummyadd.zip). When I try to install it on CentOS 7, I get the following error:
```
Running transaction
Error unpacking rpm package libdummyadd-0.1.0-1.x86_64
  Installing : libdummyadd-0.1.0-1.x86_64                                   1/1 
error: unpacking of archive failed on file /usr/lib64/libdummyadd.so;6476433e: cpio: Digest mismatch
  Verifying  : libdummyadd-0.1.0-1.x86_64                                   1/1 

Failed:
  libdummyadd.x86_64 0:0.1.0-1    
```

`rpm -v --checksig libdummyadd.rpm` returns:
```
libdummyadd.rpm:
    Header SHA1 digest: OK (77492d16c6b099e0bc74cc864fe316c7f7dfe6b6)
    MD5 digest: OK (7decf15b91517ed4fb2edc616914dbe2)
```

Running `rpm -qp --xml libdummyadd.rpm` I get:
<details> 
  <summary>Expand</summary>

```xml
<rpmHeader>
  <rpmTag name="Headeri18ntable">
	<string>C</string>
  </rpmTag>
  <rpmTag name="Sigmd5">
	<base64>CcK1Fb+YZqLtn6QTsd0GRQ==
</base64>
  </rpmTag>
  <rpmTag name="Sha1header">
	<string>f4e9822feed5995e8234a1525bbcb742955f7937</string>
  </rpmTag>
  <rpmTag name="Longsigsize">
	<integer>15401</integer>
  </rpmTag>
  <rpmTag name="Name">
	<string>libdummyadd</string>
  </rpmTag>
  <rpmTag name="Version">
	<string>0.1.0</string>
  </rpmTag>
  <rpmTag name="Release">
	<string>1</string>
  </rpmTag>
  <rpmTag name="Epoch">
	<integer>0</integer>
  </rpmTag>
  <rpmTag name="Summary">
	<string>Dummy library for testing RPM generation</string>
  </rpmTag>
  <rpmTag name="Description">
	<string>Dummy library for testing RPM generation</string>
  </rpmTag>
  <rpmTag name="Buildtime">
	<integer>1685472520</integer>
  </rpmTag>
  <rpmTag name="Buildhost">
	<string>host</string>
  </rpmTag>
[...]

PR #152, opened 2023-05-31
Include build time by default

### LINKED ISSUE #147: Potential validity issues with generated RPMs
Let's take this [dummy library](https://github.com/newpavlov/libdummyadd) as an example. By running `cargo build --release` and `cargo generate-rpm` we get `libdummyadd-0.1.0-1.x86_64.rpm`. It can be installed and uninstalled without issues. But if we run `rpm --rebuilddb` after installation, we get the following warning: `error: header #2054 in the database is bad -- skipping.`. Now the packet is not listed as installed, e.g. we can no longer remove it using packet manager. Additionally, `mc` can not open generated packets, while it handles packets generated by `rpmbuild` just fine. This behavior can be reproduced on at least CentOS 7 and Fedora 32.

This can be reproduced by using the latest version of `rpm` crate directly:
<details> 
  <summary>Expand</summary>

```rust
let pkg = rpm::RPMBuilder::new(
    "libdummyadd",
    "0.1.0",
    "UNLICENSE",
    "x86_64",
    "Dummy library for testing RPM generation",
)
.compression(rpm::CompressionType::None)
.with_file(
    "/path/to/libdummyadd/target/release/libdummyadd.so",
    rpm::RPMFileOptions::new("/usr/lib64/libdummyadd.so"),
)?
.build_time(rpm::chrono::Utc::now())
.build_host("host".to_string())
.vendor("NA")
.build()?;

let mut f = std::fs::File::create("libdummyadd.rpm")?;
pkg.write(&mut f)?;
```

</details>

If my understanding is correct, the headers added after installation of the package are considered invalid by RPM, meaning that RPMs generated by this crate may not be completely valid.

newpavlov:
Closes #147.

In a future breaking release it's probably worth to change API of `build_time` to something similar to code in this [comment](https://github.com/rpm-rs/rpm/issues/147#issuecomment-1568653374).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## Changed
+
+- Build time is now included into package by default.
+
 ### Fixed
 
 - CentOS 7 support by using long sizes only for packages bigger than 4 GiB
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -41,7 +41,6 @@
 //!                         .user("hugo"),
 //!             )?
 //!             .pre_install_script("echo preinst")
-//!             .build_time(chrono::Utc::now())
 //!             .build_host(gethostname::gethostname().to_str().expect("Hostname works. qed").to_string())
 //!             .add_changelog_entry("Max Mustermann <max@example.com> - 0.1-29", "- was awesome, eh?", chrono::DateTime::parse_from_rfc2822("Wed, 19 Apr 2023 23:16:09 GMT").expect("Date 1 is correct. qed"))
 //!             .add_changelog_entry("Charlie Yom <test2@example.com> - 0.1-28", "- yeah, it was", chrono::DateTime::parse_from_rfc3339("1996-12-19T16:39:57-08:00").expect("Date 2 is corrrect. qed"))
--- a/src/rpm/builder.rs
+++ b/src/rpm/builder.rs
@@ -90,7 +90,7 @@ pub struct RPMBuilder {
     vcs: Option<String>,
     cookie: Option<String>,
 
-    build_time: Option<chrono::DateTime<chrono::Utc>>, // because `rpm_time_t` is an `uint32`
+    build_time: Option<u32>,
     build_host: Option<String>,
 }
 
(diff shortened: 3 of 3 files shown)


dralley:
For the sake of not changing APIs multiple times let's figure out what API we want long-term and just go ahead and do that.

@drahnr The situation is basically that `RPMTAG_BUILDTIME` is pseudo-required in that plenty of software including `rpm` itself seems to expect it.  So the question is how do we balance this with the desire to make sure reproducible builds are simple to do.

  I'm still partial to my suggestion obviously.   If we want builds to be reproducible by default and also not be too problematic in terms of compatibility, my recommendation is to use the timestamp of the newest changelog by default and make it configurable.

newpavlov:
>For the sake of not changing APIs multiple times let's figure out what API we want long-term and just go ahead and do that.

This PR is a backwards compatible stop-gap solution. I think a "proper" API for this and mtimes should be done as part of v0.12. I also plan to submit a PR with basic API for overwriting mtimes, to replace `touch`-based workaround which I currently use in my project.

newpavlov:
@drahnr 
No, installation failures in #148 were fixed by #149. This PR fixes the issue with `rpm --rebuilddb`.

dralley:
Well considering the previous goal was to have reproducible builds by default, we ought to do 0.12 anyway as it in theory would be a breaking change.

(In practice I very much doubt anyone cares.  But I have another feature for 0.12 anyway)

newpavlov:
>In practice I very much doubt anyone cares.

Well, personally I do and need it for shipping the mentioned project. :)

dralley:
I mean in terms of the breakage.  But either way I'd still rather make this change in a 0.12 than a 0.11.z

newpavlov:
You mean this PR is a breaking change? I don't think so. In terms of API it's certainly not.

Considering the incompatibility with `--rebuilddb` it's a clear bug fix. The only "affected" users are those who did not use `build_time` in their code, otherwise behavior of the crate will be the same. You could argue that this change may break reproducibility for such users, but I think fixing the incompatibility should take precedence.

dralley:
>You mean this PR is a breaking change? I don't think so. In terms of API it's certainly not.

> Considering the incompatibility with --rebuilddb it's a clear bug fix.

I don't mean API breaking, I mean that adding a build time by default makes the build no longer reproducible.

PR #155, opened 2023-06-02
Add source_date method

newpavlov:
Depends on #152

Relevant issue: #117

In v0.12 it may be worth to remove the `build_time` method altogether and instead set build time to `source_date` if it was set.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## Added
+
+- `source_date` method for clamping modification time of files
+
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## Added
+
+- `source_date` method for clamping modification time of files
```
dralley, line 11:
```suggestion
- `RPMBuilder::source_date()` method for clamping modification time of files. This allows reproducible builds when included files are re-generated (having the same content but with different mtimes).
```

dralley (approved):

Issue #156, opened 2023-06-02
Sort out ergonomics of reproducable builds


dralley:
re: https://github.com/rpm-rs/rpm/issues/117

Reproducable builds are possible now, if you provide both a `build_time` and a `source_date`.   But it would be best if users didn't need to know all the knobs.  In RPM, those knobs are:

* clamping file mtimes to a "source date epoch" value
* overriding the build time with the value of  "source date epoch"
* deriving "source date epoch" from the changelogs, which are static
* use a deterministic hostname

We avoid 4 by virtue of not setting the hostname by default.  We now have the ability to perform 1.  But we force the user to provide both `source_date` and a `build_time` manually and we don't provide a changelog shortcut, so it's not as simple as it could potentially be.

1) Is it OK to consolidate the options?
    * Is it OK to assume that the user is always ok with file mtime clamping behavior?
    * Is it OK to assume that `build_time` and `source_date` (the clamp value) ought to be the same in all cases? 
2) Should we provide some way to derive `source_date` from the changelogs automatically?

dralley:
For what it's worth re: build_time `--rebuilddb` issues, they accepted as a bug and it has been fixed upstream.

newpavlov:
I have implemented the clamping behavior for both modification and build times in #157.

>Should we provide some way to derive source_date from the changelogs automatically?

I think it should be handled by a higher-level software, not by this library. IIUC it's recommended to use timestamp of last commit in the project's VCS, which often is different from last changelog entry.

>re: build_time --rebuilddb issues, they accepted as a bug and it has been fixed upstream.

Good to know! Unfortunately, there is a lot of systems in the wild with outdated `rpm`, so we still have to include the build time tag to avoid potential problems.

dralley:
I brought this up on Matrix with the person who contributed the ` %use_source_date_epoch_as_buildtime` option to RPM

> Him: I requested this flag be made because buildtime is used in some cases as a tie-break for packages of the same NEVRA in build systems and package managers. basically, it broke OBS and YUM if the buildtime was statically set

> Me: but the point of source_date_epoch is to make packages reproducable right?  If you allow the buildtime to be different, then the package is no longer reproducable, and there's no reason to use source_date_epoch. couldn't source_date_epoch just be disabled instead

> Him: I didn't want to say it outright in the commit, but it's quite common that build-time is used as a tie-breaker

[...]

PR #157, opened 2023-06-02
Remove the build_time method, use source_date for clamping all package times

newpavlov:
The package build time is now gets clamped by date provided to the `source_date` method.


src/rpm/builder.rs
```diff
@@ -35,15 +36,20 @@ fn file_mode(_file: &fs::File) -> Result<u32, RPMError> {
...
 
-fn system_time_as_u32(when: std::time::SystemTime) -> u32 {
-    let dt = chrono::DateTime::<chrono::Utc>::from(when);
-    date_time_as_u32(&dt)
+fn system_time_as_u32(dt: std::time::SystemTime) -> u32 {
+    dt.duration_since(SystemTime::UNIX_EPOCH)
+        .expect(PAST_ERROR)
+        .as_secs()
+        .try_into()
+        .expect(FUTURE_ERROR)
```
newpavlov, line 52: AFAIK we do not need `chrono` for converting `SytemTime` into UNIX timestamp, i.e. `std` should handle host time zone properly by itself.

dralley (approved):
@drahnr @cmeister2 Please evaluate in the context of https://github.com/rpm-rs/rpm/issues/156, do you think it's OK to squash `build_time` and `source_date` into one option?

CHANGELOG.md
```diff
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Added
 
-- `RPMBuilder::source_date()` method for clamping modification time of files. This allows reproducible builds when included files are re-generated (having the same content but with different mtimes).
+- `RPMBuilder::source_date_epoch()` method for clamping modification time of files and build time of the package. This allows reproducible builds when included files are re-generated (having the same content but with different mtimes).
```
dralley, line 12: Change this to `Breaking Change` and explain that `build_time` as a solitary option is going away

newpavlov: I've marked removal of `build_time` as a breaking change and kept `source_date_epoch` as an addition.

dralley (changes requested):
Awaiting consensus on https://github.com/rpm-rs/rpm/issues/156#issuecomment-1575693663

newpavlov:
@dralley 
Any updates? IIUC you've agreed in the issue that we need proper bitwise reproducible builds.

src/rpm/builder.rs
```diff
@@ -259,31 +242,24 @@ impl RPMBuilder {
...
-        datetime: chrono::DateTime<TZ>,
-    ) -> Self
-    where
-        N: AsRef<str>,
-        E: AsRef<str>,
-        TZ: chrono::TimeZone,
-    {
+        name: impl AsRef<str>,
+        entry: impl AsRef<str>,
+        timestamp: impl TryInto<Timestamp, Error = impl Debug>,
```
dralley, line 251: @drahnr Can you live with this?

CHANGELOG.md
```diff
@@ -6,17 +6,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+## Added
 
-- Build time is now included into package by default
+- `RPMBuilder::source_date_epoch` method for clamping modification time of files, build time of the package, and signature time. This functionality is required for reproducible generation of packages.
 
 ### Fixed
 
-- CentOS 7 support by using long sizes only for packages bigger than 4 GiB
+- CentOS 7 support by using long sizes only for packages bigger than 4 GiB.
+- Compatibility with `rpm --rebuilddb` by including package build time by default.
```
dralley, line 21: Package build time still deserves an entry under `Added`

src/rpm/package.rs
```diff
@@ -101,7 +99,7 @@ impl RPMPackage {
 
     /// Create package signatures using an external key and add them to the signature header
     #[cfg(feature = "signature-meta")]
-    pub fn sign<S>(&mut self, signer: S) -> Result<(), RPMError>
+    pub fn sign<S>(&mut self, signer: S, t: Timestamp) -> Result<(), RPMError>
```
dralley, line 102: Add an example here

dralley:
`build_and_sign()` ought to use the `source_date_epoch` timestamp during signing

newpavlov:
>build_and_sign() ought to use the source_date_epoch timestamp during signing

It already [does](https://github.com/rpm-rs/rpm/blob/a8ff7250aba46b8f193b88cccd3b4757cf3028ae/src/rpm/builder.rs#L528-L533), no? Or have I missed something?

CHANGELOG.md
newpavlov: Do you want both lines:
```md
### Added
- Package build time is now included by default.
### Fixed
- Compatibility with `rpm --rebuilddb` by including package build time by default.
```
?

Or only the former?

dralley:
I missed that part.  In that case can we just replace `sde` with `self.source_date_epoch`?  It's much clearer.

src/rpm/package.rs
newpavlov: Hm, FYI none of the other `RPMPackage` methods contain examples. Should I construct `RPMPacakge` in example using the builder?

CHANGELOG.md
dralley: Both, but the second line could be something more like 

> Avoid a longstanding bug in `rpm --rebuilddb` by adding a package build time by default.

dralley:
Also, if we're passing our own object rather than "just" an epoch value, `source_date` might be better honestly.

src/rpm/package.rs
dralley: You can just copy a minimal part of the example, just enough to demonstrate signing and the various ways you could construct a `Timestamp`.

newpavlov:
> In that case can we just replace sde with self.source_date_epoch?
>
> if we're passing our own object rather than "just" an epoch value, source_date might be better honestly.

Done.

src/rpm/package.rs
dralley: Also maybe instead of changing the entire signature it could be two functions, where `sign()` just calls `sign_with_timestamp(key, Timestamp::now())`

newpavlov: Done.

dralley (approved):
See one last comment about changelog

PR #159, opened 2023-06-14
Remove RPM prefix from type names

newpavlov:
Closes #158

PR #160, opened 2023-06-16
Add RPMTAG_RPMVERSION tag to headers

dralley:
Purpose: Allows you to debug which library, and which version of the library the package was built with

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,13 +8,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Breaking Change
 
-- The `RPMBuilder::build_time` method is removed. Package build time is now included by default and can be clamped using the `RPMBuilder::source_date` method.
+- The `RPMBuilder::build_time` method is removed. Package build time is now included by default
+  and can be clamped using the `RPMBuilder::source_date` method.
+- Several of the signer and verifier trait APIs were changed
 
 ## Added
 
 - `RPMBuilder::source_date` method for clamping modification time of files, build time of the package, and signature time. This functionality is required for reproducible generation of packages.
 - `RPMPackage::sign_with_timestamp` method.
-- Package build time is now included by default.
+- The "rpmversion" tag is now populated so that packages know which library (and version)
+  they were built with.
+
+### Changed
+
+- Build time metadata is now included in the built package by default
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


src/rpm/builder.rs
```diff
@@ -762,7 +762,11 @@ impl RPMBuilder {
                 offset,
                 IndexData::Int32(vec![self.epoch]),
             ),
-            // @todo: write RPMTAG_RPMVERSION?
+            IndexEntry::new(
+                IndexTag::RPMTAG_RPMVERSION,
+                offset,
+                IndexData::StringTag(format!("rpm-rs {}", env!("CARGO_PKG_VERSION"))),
```
newpavlov, line 768: What about tools like `cargo-generate-rpm`? Maybe it's worth to make it configurable, with `rpm-rs ...` being the default value?

Also, instead of `format!` you could use `concat!("rpm-rs ", env!("CARGO_PKG_VERSION"))`.

dralley: It shouldn't be configurable because the part that actually matters for debugging purposes is knowing what version of this library the package was created by.  `cargo-generate-rpm` provides the metadata, but that's a very foolproof process compared to constructing a package from it.

CHANGELOG.md
```diff
@@ -8,13 +8,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Breaking Change
 
-- The `RPMBuilder::build_time` method is removed. Package build time is now included by default and can be clamped using the `RPMBuilder::source_date` method.
+- The `RPMBuilder::build_time` method is removed. Package build time is now included by default
+  and can be clamped using the `RPMBuilder::source_date` method.
```
newpavlov, line 12: I have fixed line length in #159, so you can remove this change.

dralley:
@drahnr @cmeister2 can one of you review this?

cmeister2 (commented):
Happy with this - is it possible/sensible to add a unit test for this?

dralley:
It's possible but I'm not sure how specific we would want the unit test to be.  For instance, we could just check for the existence of the tag, or we could check the value of the tag, but the value would need to be updated for each new release.

I'd be OK with either I think.  The latter might be annoying, but actually it's what `rpm` upstream pretty much does.  They have reproducibility tests within versions, and the expected checksums need to be updated after every release.

drahnr (approved):

PR #164, opened 2023-07-04
Set RPMTAG_SOURCERPM for all packages

### LINKED ISSUE #163: SOURCERPM rpm tag
We found an problem in our internal CI system when, in centos 7, a RPM package is generated with cargo-generate-rpm but it is signed externally using `rpmsign` in a debian system (that is isolate to protect the signing key).

The signed rpm package can't be installed in centos 7 using `rpm -i` and produce this error:
```
error: source package contains no .spec file
error: /tmp/hello/target/generate-rpm/hello-0.1.0-1.x86_64.rpm cannot be installed
```

<details>
<summary>See these commands to reproduce the problem</summary>

```sh
docker run --rm -i -v /tmp:/tmp -w /tmp  centos:7 << EOF_DOCKER_CENTOS7
# install c/c++ compiler...
yum group install -y "Development Tools"
# install rust
curl --proto '=https' --tlsv1.2 https://sh.rustup.rs -sSf | sh -s -- --profile minimal -y
source /root/.cargo/env
# build minimal rust executable
cargo new hello
cd hello
cargo build --release
# generate rpm by using cargo-generate-rpm 
# use version 0.11 to avoid https://github.com/rpm-rs/rpm/issues/148 (already fixed but in unreleased version)
cargo install --version 0.11.0 cargo-generate-rpm
cat >> Cargo.toml << EOF
[package.metadata.generate-rpm]
assets = [
    { source="target/release/hello",  dest="/usr/local/bin/", mode="755" },
]
license = "MIT"
description = "Hello world"
EOF
cargo generate-rpm --payload-compress=gzip
EOF_DOCKER_CENTOS7

docker run --rm -i -v /tmp:/tmp -w /tmp  debian:11 << EOF_DOCKER_DEBIAN11
apt-get update
apt-get install -y gnupg2 rpm
gpg  --batch --pinentry-mode loopback --passphrase '' --quick-generate-key GPG_NAME_rpm_signer
echo %_gpg_name GPG_NAME_rpm_signer > ~/.rpmmacros
rpmsign --resign /tmp/hello/target/generate-rpm/hello-0.1.0-1.x86_64.rpm
EOF_DOCKER_DEBIAN11

docker run --rm -i -v /tmp:/tmp -w /tmp  centos:7 << EOF_DOCKER_CENTOS7
rpm -vv -i /tmp/hello/target/generate-rpm/hello-0.1.0-1.x86_64.rpm
EOF_DOCKER_CENTOS7
```

</details>
<br>

It seems that the `rpmsign` convert the generated binary rpm package to a source rpm. Only the old `rpm` version in centos 7 is confused by this, as the signed package can be installed in more modern linux distributions.

The `rpmsign` thinks that is processing a source RPM and produce a source RPM because it seems that the presence of the SOURCERPM tag is used by the RPM library to detect if a RPM package is a binary RPM or a source RPM (SRPM), see:

- https://stackoverflow.com/questions/12585927/how-to-a-detect-if-a-rpm-is-a-source-rpm-or-a-binary-rpm

- https://github.com/rpm-software-management/rpm/blob/7b5db7d98e024ebc582242712e56e603616144c0/lib/headerutil.c#L67
  ```c
  int headerIsSource(Header h)
  {
    return (!headerIsEntry(h, RPMTAG_SOURCERPM));
  }
  ```
- https://github.com/rpm-software-management/rpm/blob/master/lib/rpmlead.c#L61
  ```c
  static int rpmLeadFromHeader(Header h, struct rpmlead_s *l)
  {
    // ...
	l->type = (headerIsSource(h) ? 1 : 0);
    // ...
  }
  ```

This crate doesn't add the SOURCERPM tag but maybe it should be.

ibasurto:
### 📜 Checklist

Close #163


src/rpm/builder.rs
```diff
@@ -747,6 +747,14 @@ impl RPMBuilder {
         let small_package = combined_file_sizes <= u32::MAX.into();
 
         let mut actual_records = vec![
+            IndexEntry::new(
+                IndexTag::RPMTAG_SOURCERPM,
+                offset,
+                IndexData::StringTag(format!(
+                    "{}-{}-{}.nosrc.rpm",
```
cmeister2, line 754: Looking at RPM, this gets set to "(none)" by RPM internally when no headers are present. https://github.com/rpm-software-management/rpm/blob/master/lib/package.c#L284C1-L284C53

Maybe just using "(none)" is better?

cmeister2:
Given we're not targeting SRPMs with this library I expect this is fine to set. I'll let another maintainer have a look as well though. Do make sure to go through the checklist too - CHANGELOG, unit test, etc.

src/rpm/builder.rs
ibasurto: I also think now that "(none)" is better, thanks!

src/tests.rs
```diff
@@ -39,6 +39,8 @@ fn test_rpm_builder() -> Result<(), Box<dyn std::error::Error>> {
 
     pkg.write(&mut buff)?;
 
+    pkg.metadata.get_source_rpm()?;
```
dralley, line 42: Leave a comment about why this is here.  Otherwise, looks fine.

dralley (changes requested):

src/tests.rs
dralley: Please squash the 3 commits into one, also.

dralley:
Eventually this ought to be configurable, that is, `RpmBuilder` ought to allow you to define the value with a method.

But I don't see that s being something that needs to be done immediately.

src/rpm/builder.rs
ibasurto: Done!

dralley (approved):

ibasurto:
Thanks @dralley and @cmeister2 !

PR #166, opened 2023-08-15
Add support for setting RPMTAG_FILECAPS header

dsteeley:
This change adds support for setting the RPMTAG_FILECAPS header to add capabilities to the files packaged in the rpm. 

### 📜 Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,7 @@ let pkg = rpm::PackageBuilder::new("test", "1.0.0", "MIT", "x86_64", "some aweso
         // you can set a custom mode and custom user too
         rpm::FileOptions::new("/etc/awesome/second.toml")
             .mode(rpm::FileMode::regular(0o644))
+            .caps("cap_sys_admin,cap_net_admin=pe")
             .user("hugo"),
     )?
     .pre_install_script("echo preinst")
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ let pkg = rpm::PackageBuilder::new("test", "1.0.0", "MIT", "x86_64", "some aweso
         // you can set a custom mode and custom user too
         rpm::FileOptions::new("/etc/awesome/second.toml")
             .mode(rpm::FileMode::regular(0o644))
-            .caps("cap_sys_admin,cap_net_admin=pe")
+            .caps("cap_sys_admin,cap_net_admin=pe")?
             .user("hugo"),
     )?
     .pre_install_script("echo preinst")
--- a/src/rpm/builder.rs
+++ b/src/rpm/builder.rs
@@ -898,11 +898,6 @@ impl PackageBuilder {
                     IndexData::Int16(file_modes),
                 ),
                 IndexEntry::new(
-                    IndexTag::RPMTAG_FILECAPS,
-                    offset,
-                    IndexData::StringArray(file_caps),
-                ),
-                IndexEntry::new(
                     IndexTag::RPMTAG_FILERDEVS,
                     offset,
                     IndexData::Int16(file_rdevs),
@@ -978,6 +973,13 @@ impl PackageBuilder {
                     IndexData::StringArray(self.directories.into_iter().collect()),
                 ),
             ]);
+            if file_caps.iter().any(|caps| !caps.is_empty()) {
+                actual_records.extend([IndexEntry::new(
+                    IndexTag::RPMTAG_FILECAPS,
+                    offset,
+                    IndexData::StringArray(file_caps),
+                )])
+            }
         }
 
         actual_records.extend([
--- a/src/rpm/builder.rs
+++ b/src/rpm/builder.rs
@@ -980,7 +980,12 @@ impl PackageBuilder {
                 actual_records.extend([IndexEntry::new(
                     IndexTag::RPMTAG_FILECAPS,
                     offset,
-                    IndexData::StringArray(file_caps.iter().map(|f| f.unwrap().to_string()).collect::<Vec<String>>()),
+                    IndexData::StringArray(
+                        file_caps
+                            .iter()
+                            .map(|f| f.unwrap().to_string())
+                            .collect::<Vec<String>>(),
+                    ),
                 )])
             }
         }
--- a/src/rpm/builder.rs
+++ b/src/rpm/builder.rs
@@ -983,7 +983,10 @@ impl PackageBuilder {
                     IndexData::StringArray(
                         file_caps
                             .iter()
-                            .map(|f| f.unwrap_or(capctl::FileCaps::empty()).to_string())
+                            .map(|f| match f {
+                                Some(caps) => caps.to_string(),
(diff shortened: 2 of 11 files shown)


dralley:
It would be great if we could fill out support for reading filecaps also, so that we can write functional tests for this.

dralley:
I haven't checked the behavior of RPM, but if it doesn't set FILECAPS if none were used in the specfile, we should probably avoid setting that tag if the user didn't provide any as well.

If it always sets the FILECAPS tag, then this is fine.

dsteeley:
I've got these changes on a separate branch at the moment. 
Would you like them here with a functional test or kept separate? 

My understanding of the behaviour [here](https://github.com/rpm-software-management/rpm/blob/f9775b454e1969a8d5c8d8b7435817876ed79482/build/files.c#L1238) is that if not present then the header isn't added. 
I've updated the behaviour to not add the header unless any file has a capability.

README.md
```diff
@@ -50,6 +50,7 @@ let pkg = rpm::PackageBuilder::new("test", "1.0.0", "MIT", "x86_64", "some aweso
         // you can set a custom mode and custom user too
         rpm::FileOptions::new("/etc/awesome/second.toml")
             .mode(rpm::FileMode::regular(0o644))
+            .caps("cap_sys_admin,cap_net_admin=pe")
```
dralley, line 53: Is there any validation that can or should be done here?  (Genuinely a question, it might not be worth it)

src/rpm/package.rs
```diff
@@ -849,15 +867,20 @@ impl PackageMetadata {
...
+                    |mut acc, (path, user, group, mode, digest, mtime, size, flags, _cap)| {
                         let digest = if digest.is_empty() {
                             None
                         } else {
                             Some(FileDigest::load_from_str(algorithm, digest)?)
                         };
+                        let cap = match caps {
+                            Some(ref caps) => caps.get(acc.len()).map(|x| x.to_owned()),
+                            None => None
+                        };
```
dralley, line 883: I suppose this works, but do you think indexing directly would be cleaner? i.e. drop changes on line 870 and 874, use enumerate on the iterator to get a direct index, and fetch the index from `caps`

I think it would be a little more complicated than that but that's the basic idea.  The reason I suggest it is that there's probably a few other bits of metadata that would work this way, and that approach feels potentially easier to scale than this one.

README.md
dsteeley: Looked into the options, RPM verifies by calling libcaps `cap_from_text()`. 
There are several crates providing this mechanism I've gone with `capctl` as it provides a FileCaps struct that will verify the string provided by a user. 

I then convert back to String for the writing of the header, hope that's an acceptable solution?

dsteeley: I've looked into the validation that [RPM does](https://github.com/rpm-software-management/rpm/blob/44c76481e75342b6f9e1994d1bb7f83575341c55/lib/rpmfi.c#L789). 
I don't think that's appropriate to add here as it's only verifying that some files have been provided

dralley:
There are some test failures but apart from that I like the way this looks.  

Once that's addressed go ahead and squash it all into one commit (or I can do that when merging it, I suppose)

README.md
```diff
@@ -50,7 +50,7 @@ let pkg = rpm::PackageBuilder::new("test", "1.0.0", "MIT", "x86_64", "some aweso
         // you can set a custom mode and custom user too
         rpm::FileOptions::new("/etc/awesome/second.toml")
             .mode(rpm::FileMode::regular(0o644))
-            .caps("cap_sys_admin,cap_net_admin=pe")
+            .caps("cap_sys_admin,cap_net_admin=pe")?
```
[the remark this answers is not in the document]
dsteeley: Correct, I think the resulting form is the result of passing the string to and from `capctl`. 

It uses the libcap implementations, https://docs.rs/capctl/latest/capctl/caps/struct.FileCaps.html. Which I think normalize input.

src/rpm/builder.rs
```diff
@@ -983,7 +983,7 @@ impl PackageBuilder {
                     IndexData::StringArray(
                         file_caps
                             .iter()
-                            .map(|f| f.unwrap().to_string())
+                            .map(|f| f.unwrap_or(capctl::FileCaps::empty()).to_string())
```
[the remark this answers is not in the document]
dralley: It looks like they just use empty strings

["", "cap_net_admin=ep", "", "cap_net_admin=ep", "", "", ""]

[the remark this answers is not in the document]
dsteeley: I've updated so that no capability is mapped to an empty string.

src/tests.rs
```diff
@@ -45,6 +46,23 @@ fn test_rpm_builder() -> Result<(), Box<dyn std::error::Error>> {
...
+    // check various metadata on the files
+    pkg.metadata.get_file_entries()?.iter().for_each(|f| {
+        if f.path.as_os_str() == "/etc/awesome/second.toml" {
+            assert_eq!(
+                f.clone().caps.unwrap(),
+                "cap_sys_ptrace,cap_sys_admin=ep".to_string()
+            );
+            assert_eq!(f.ownership.user, "hugo".to_string());
+        } else if f.path.as_os_str() == "/etc/awesome/config.toml" {
+            // The "=" capability is equivalent to an empty capability set
```
dralley, line 58: Not sure if this comment is still relevant, but also it's not harmful

dralley (approved):

dralley:
>The failing job is an issue compiling time-macros on rust 1.65.  Any preference or suggestion on how I resolve?

You can just pin `time` to an older (compatible) version for now.

src/tests.rs
dsteeley: Removed

dralley:
@dsteeley Are you looking for a release soon or is waiting a few weeks OK?

 And just out of curiosity, are you actively using (or considering) this for building RPMs at MS?

dsteeley:
> @dsteeley Are you looking for a release soon or is waiting a few weeks OK?
> 
> And just out of curiosity, are you actively using (or considering) this for building RPMs at MS?

Not in a rush, although I'd like to get this plumbed through cargo-generate-rpm at some point. 

We are already using this (via cargo-generate-rpm) to package rust binaries as rpms and then containerise using [rpmoci](https://github.com/microsoft/rpmoci).
I'd like to have to option to set capabilities on those binaries, currently doing it via post install script which isn't the neatest solution.

dsteeley:
@dralley Are you planning to squash this or do you want me to combine into one commit?

dralley:
Nah I got it.

dralley:
@dsteeley Thanks!

README.md
dralley: @dsteeley FYI, I think I will make this into a feature (probably a default one but I'm not certain yet) to restore the ability to compile on Windows

PR #167, opened 2023-08-15
DRAFT: Add capability parsing

dsteeley:
### 📜 Checklist

### DIFF
--- a/src/rpm/package.rs
+++ b/src/rpm/package.rs
@@ -833,10 +833,10 @@ impl PackageMetadata {
             .header
             .get_entry_data_as_u32_array(IndexTag::RPMTAG_FILEFLAGS);
         // @todo
-        // let caps = self.get_entry_i32_array_data(IndexTag::RPMTAG_FILECAPS)?;
+        let caps = self.header.get_entry_data_as_string_array(IndexTag::RPMTAG_FILECAPS);
 
-        match (modes, users, groups, digests, mtimes, sizes, flags) {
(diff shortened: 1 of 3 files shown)


src/rpm/package.rs
```diff
@@ -833,10 +833,10 @@ impl PackageMetadata {
             .header
             .get_entry_data_as_u32_array(IndexTag::RPMTAG_FILEFLAGS);
         // @todo
-        // let caps = self.get_entry_i32_array_data(IndexTag::RPMTAG_FILECAPS)?;
+        let caps = self.header.get_entry_data_as_string_array(IndexTag::RPMTAG_FILECAPS);
 
-        match (modes, users, groups, digests, mtimes, sizes, flags) {
-            (Ok(modes), Ok(users), Ok(groups), Ok(digests), Ok(mtimes), Ok(sizes), Ok(flags)) => {
+        match (modes, users, groups, digests, mtimes, sizes, flags, caps) {
+            (Ok(modes), Ok(users), Ok(groups), Ok(digests), Ok(mtimes), Ok(sizes), Ok(flags), Ok(caps)) => {
```
dralley, line 839: This will need to take into account the (very likely) possibility that there won't be `caps` to iterate over, even when the other tags exist.  Probably we need to handle this particular tag independently of the rest.

dsteeley: I've moved this to the other capabilities PR. 
To solve this issue I've gone with representing `caps` as a `Result<Option<&[String]>>`. 

If the header isn't found then the assumption is made that no files have capabilities set and the caps field is `None`

PR #170, opened 2023-08-25
Improve file mode API

dralley:
### 📜 Checklist


dralley:
@syokensyo Here's my unfinished first draft of the API redesign I mentioned.  Do you have any objections to the overall direction of this?

src/rpm/headers/types.rs
```diff
@@ -32,30 +32,35 @@ pub struct PackageFileEntry {
...
 
-// @todo: <https://github.com/rpm-rs/rpm/issues/52>
-impl From<u16> for FileMode {
-    fn from(raw_mode: u16) -> Self {
+const REGULAR_FILE_TYPE: u16 = 0o100000; //  bit representation = "1000000000000000"
+const DIR_FILE_TYPE: u16 = 0o040000; //      bit representation = "0100000000000000"
+const SYMBOLIC_LINK_FILE_TYPE: u16 = 0o120000; // bit representation = "1010000000000000"
+
+// @todo: turn permissions into a defined bitset
+// @todo: resolve ambiguity between mode and symlink - .symlink() ought to set the mode
```
dralley, line 58: Maybe instead of `FileOptions::new()` we should have `FileOptions::regular(dest)`, `FileOptions::dir(dest)`, and `FileOptions::symlink(dest, link)` just like `FileMode`, and drop `.mode()`  on `FileOptionsBuilder` in favor of `.permissions()`

That would resolve the ambiguity.

syokensyo: That's more clear than before, and for this `fn try_from(raw_mode: i32)`  is is possible to change `i32` to `u16`? so we can no need the runtime value check.

README.md
```diff
@@ -38,21 +38,25 @@ let pkg = rpm::PackageBuilder::new("test", "1.0.0", "MIT", "x86_64", "some aweso
...
+        rpm::FileOptions::regular("/etc/awesome/second.toml")
+            .permissions(0o644)
             .caps("cap_sys_admin,cap_net_admin=pe")?
             .user("hugo"),
     )?
+    .with_file(
+        "./test_assets/empty_file_for_symlink_create",
+        rpm::FileOptions::symlink("/usr/bin/awesome_link", "/usr/bin/awesome")
+            .permissions(0o644)
     .pre_install_script("echo preinst")
```
dralley, line 60: @drahnr Could you do a quick skim of these changes and let me know what you think?  No need to dig into it, it still needs polishing.

src/rpm/builder.rs
```diff
@@ -264,17 +264,17 @@ impl PackageBuilder {
...
     ///     // file mode is inherited from source file
     ///     .with_file(
     ///         "./awesome-bin",
-    ///         rpm::FileOptions::new("/usr/bin/awesome"),
+    ///         rpm::FileOptions::regular("/usr/bin/awesome"),
     ///     )?
     ///      .with_file(
     ///         "./awesome-config.toml",
     ///         // you can set a custom mode, capabilities and custom user too
-    ///         rpm::FileOptions::new("/etc/awesome/second.toml").mode(0o100744).caps("cap_sys_admin=pe")?.user("hugo"),
```
dralley, line 277: Setting the full "mode" manually would be a massive footgun and as far as I can tell RPM doesn't document that you can do it nor do any of their test specfiles.  They say you can set the "mode" with `%attr()` but actually it seems to only be the permissions.

src/rpm/headers/types.rs
```diff
@@ -32,30 +32,32 @@ pub struct PackageFileEntry {
...
 #[derive(Copy, Clone, PartialEq, Eq, Debug, Hash)]
-pub enum FileMode {
-    // It does not really matter if we use u16 or i16 since all we care about
-    // is the bit representation which is the same for both.
-    Dir { permissions: u16 },
-    Regular { permissions: u16 },
-    SymbolicLink { permissions: u16 },
-    // For "Invalid" we use a larger integer since it is possible to create an invalid
-    // FileMode by providing an overflowing integer.
-    Invalid { raw_mode: i32, reason: &'static str },
```
dralley, line 45: I don't think this should be one of our "valid" states, we can just return an error.  If we want to parse invalid RPMs then we can compute the mode lazily.

src/rpm/headers/types.rs
```diff
@@ -217,35 +217,60 @@ impl FileOptions {
...
 #[derive(Debug)]
 pub struct FileOptionsBuilder {
     inner: FileOptions,
 }
 
+// @todo: finish support for different types of file attributes (rpmfileAttrs)
+// see: constants::FileFlags
+// @todo: should we represent "defattr", that is, set default permissions on all files in a package
+// without needing to explicitly them for each FileOptions
+// @todo: how about "%docdir"?  which automatically marks subsequent files in those directories as docs
```
dralley, line 248: open questions

src/rpm/headers/types.rs
```diff
@@ -201,9 +173,37 @@ pub struct FileOptions {
...
+    ///
+    /// Used to explicitly own the directory itself but not it's contents. This does NOT
+    /// process any of the contents of a directory.
     #[allow(clippy::new_ret_no_self)]
-    pub fn new(dest: impl Into<String>) -> FileOptionsBuilder {
+    pub fn dir(dest: impl Into<String>) -> FileOptionsBuilder {
+        // @todo: is this problematic? `with_file` has a source argument, but I do not believe %dir uses
+        // any "source" necessarily, just a destination.  Maybe we need `PackageBuilder::with_dir()`,
+        // except that sounds like it would process files recursively inside a directory, and we would
+        // still need to duplicate much of this code.
```
dralley, line 205: open questions

src/rpm/headers/types.rs
```diff
@@ -201,9 +173,37 @@ pub struct FileOptions {
     pub(crate) caps: Option<FileCaps>,
 }
 
+// @todo: should we even have "default permissions" / mode, or use Option<FileMode>?
+// if they can be skipped (unsure), 'inherit_permissions' could go away
```
dralley, line 177: open questions

src/rpm/headers/types.rs
```diff
@@ -68,108 +70,81 @@ impl From<u16> for FileMode {
...
-            }
+            })
         } else {
-            FileMode::from(raw_mode as u16)
+            Ok(FileMode::try_from(raw_mode as u16)?)
         }
     }
 }
 
+// @todo: 0o7777? should it be 0o777?
```
dralley, line 107: Apparently this is legit

> 
>  Let's fix up cdplayer with a liberal sprinkling of %attrs. Here's what the %files list looks like after we've had our way with it:
> 
>> %files
>> %attr(-, root, root) %doc README
>> %attr(4755, root, root) /usr/local/bin/cdp
>> %attr(-, root, root) /usr/local/bin/cdplay
>> %attr(-, root, rot) /usr/local/man/man1/cdp.1
>           
> 
> A couple points are worth noting here. The line for README shows that multiple macros can be used on a line — in this case, one to set file attributes, and one to mark the file as being documentation. The %attr for /usr/local/bin/cdp declares the file to be setuid root. If it sends a shiver down your spine to know that anybody can create a package that will run setuid root when installed on your system — Good! Just because RPM makes it easy to install software doesn't mean that you should blindly install every package you find.
> 
> A single RPM command can quickly point out areas of potential problems and should be issued on any package file whose creators you don't trust:
> 
>> % rpm -qlvp ../RPMS/i386/cdplayer-1.0-1.i386.rpm
>> drwxr-xr-x-  root  root   1024 Sep 13 20:16 /usr/doc/cdplayer-1.0-1
>> -rw-r--r---  root  root   1085 Nov 10 01:10 /usr/doc/cdplayer-1.0-1/README
>> -rwsr-xr-x-  root  root  40739 Sep 13 21:32 /usr/local/bin/cdp
>> lrwxrwxrwx-  root  root     47 Sep 13 21:32 /usr/local/bin/cdplay -> ./cdp
>> -rwxr-xr-x-  root   rot   4550 Sep 13 21:32 /usr/local/man/man1/cdp.1
>> % 

This almost makes me want to push the setuid/setgid bits into their own methods, but that would be a real pain for anyone trying to implement something that builds from spec files.

src/rpm/headers/types.rs
dralley: On second thought, in case we don't "inherit permissions" and there is no default, it would probably need to stat the directory to get those permissions on its own?

src/rpm/headers/types.rs
```diff
@@ -201,9 +173,37 @@ pub struct FileOptions {
...
+            },
+        }
+    }
+
+    /// Create a new FileOptions for a directory
+    ///
+    /// Used to explicitly own the directory itself but not it's contents. This does NOT
+    /// process any of the contents of a directory.
     #[allow(clippy::new_ret_no_self)]
-    pub fn new(dest: impl Into<String>) -> FileOptionsBuilder {
```
dralley, line 206: The basic principle behind the change is that symlinks need to include the destination where they link to, and the initial PR had the user provide that information with a `.symlink()`, which created ambiguity.

However, on second thought, maybe instead of having the user specify this, we should assume the file being pointed to always exists (in practice maybe we do anyway) and use `std::fs::read_link` to get that info?

dralley: @syokensyo ^ wdyt, since this impacts how we handle symlinks

syokensyo: I think it is better to make sure the `link` exists in the build root, and use `std::fs::read_link` read the `link`'s value as the target, however it is also better to provide a custom `link target` api with high priority over `link`'s value.

dralley: Leaving notes for myself

* yes `rpm` always tries to look for the files and get `stat` info, but in the case of `%ghost` files they aren't required to exist and they literally create fake stat data
* they use `lstat` which I believe is equivalent to https://doc.rust-lang.org/std/fs/fn.symlink_metadata.html
* I think `%attr` can't be applied to symlinks?

see rpm/build/files.c

Does it really make sense to provide a link override though?  If all the other files have to exist, why make a totally separate codepath?

syokensyo: "provide a link override" is flow original behavior... and yes, you are right, there is no scenario for that kind of behavior if other files exists.

PR #172, opened 2023-08-29
Fix compile error on windows

syokensyo:
### PR Content Desc
- Compile on Windows will fail because of file capabilities feature introduced, even thought rpm is specific for Linux, but there are may be some cross platform tools build on top of this lib(such as my developing tools for rpm package info read and display)

### 📜 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   of a `description`. A new method `PackageBuilder::description` can be used to
   set a detailed description for a package; if not set, the description defaults
   to the `summary`.
+- Fix compile error on Windows which introduced by file capabilities support feature.
 
 ## 0.12.1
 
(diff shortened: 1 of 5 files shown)


dralley:
@syokensyo Do you plan to undraft this?

syokensyo:
> @syokensyo Do you plan to undraft this?

There are some file conflicts, I'll resolve it first and then submit it again.

dralley:
@syokensyo Could you rebase and squash the commits?

dralley:
What do you think about simply disabling the validation on Windows rather than removing the whole feature?

syokensyo:
Yes, and capctl is platform specific, we also need some changes on it to make it work on Windows, I think that beyond the scope.

dralley:
Ideally, the functionality for parsing capability info could be split off of the rest of the `capctl` crate, whether by feature or by splitting the actual crate, and then we could use that without pulling in the unix-specific bits.  I'm not 100% certain but from a brief skim it looks possible.  But, like you said, out of scope.

I'm thinking it is perhaps better to just store a string, and do validation via `capctl` on unix platforms where it is possible.  I don't see a great reason to lock off the feature entirely.

dralley:
I've filed https://github.com/cptpcrd/capctl/issues/5#issue-1966380251

dralley:
No longer an issue

PR #174, opened 2023-08-31
Sync 0.12.1 release

dralley:
N/A

The release has already happened on a different branch


dralley:
Clippy disliked some of the verification code due to a new lint that was introduced.

I've refactored it but changed  the logging a bit in the process.  Let me know if this is wrong

drahnr (approved):

PR #178, opened 2023-10-05
PackageBuilder: both summary and description

olivierlemasle:
Currently, when building a package with `PackageBuilder`, the same string `desc` is used for both `Summary` and `Description`:

https://github.com/rpm-rs/rpm/blob/bb5c1ea3b598ebd3e03969508057ab9e1d7ab359/src/rpm/builder.rs#L794-L802

However:
- `Summary` is meant to be a _"brief, one-line summary of the package"_ (rpmlint requires a short string which does not end with a dot).
- `Description` should be a  _"full description of the software packaged in the RPM. This description can span multiple lines and can be broken into paragraphs."_ And rpmlint requires a string which does end with a dot.

In this PR, I rename the last parameter of `PackageBuilder::new()` from `desc` to `summary`, and I add a setter `description` to the package builder.

The default value for `Description` is the summary, if no description has been set. That way, this is not really a breaking change. (the default value for the description could have be the summary followed by a dot, to comply with rpmlint recommended style, but this would have a be breaking change, for a small benefit).

I kept the summary mandatory and set the description optional, because all examples in the library used for `desc` a short description which fits more the role of a summary.

### 📜 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Support for symbolic link in file mode.
 - Make file type const `REGULAR_FILE_TYPE` `DIR_FILE_TYPE` `SYMBOLIC_LINK_FILE_TYPE` public, because `FileMode::file_type` is public, sometimes we need this const to determin file type.
+- Method `PackageBuilder::new` now takes a `summary` as last parameter, instead
+  of a `description`. A new method `PackageBuilder::description` can be used to
+  set a detailed description for a package; if not set, the description defaults
+  to the `summary`.
 
 ## 0.12.1
 
--- a/src/rpm/builder.rs
+++ b/src/rpm/builder.rs
@@ -46,7 +46,8 @@ pub struct PackageBuilder {
     arch: String,
     uid: Option<u32>, // @todo: nothing is actually setting these or allowing setting them, they fall back to default
     gid: Option<u32>,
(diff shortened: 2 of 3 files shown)


src/rpm/builder.rs
```diff
@@ -794,12 +801,12 @@ impl PackageBuilder {
             IndexEntry::new(
                 IndexTag::RPMTAG_DESCRIPTION,
                 offset,
-                IndexData::I18NString(vec![self.desc.clone()]),
+                IndexData::I18NString(vec![self.desc.unwrap_or_else(|| self.summary.clone())]),
```
dralley, line 804: Works for me!

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Support for symbolic link in file mode.
 - Make file type const `REGULAR_FILE_TYPE` `DIR_FILE_TYPE` `SYMBOLIC_LINK_FILE_TYPE` public, because `FileMode::file_type` is public, sometimes we need this const to determin file type.
+- Support for setting to a package a description different from the summary.
```
dralley, line 14: The semantic signature of `PackageBuilder::new()` has changed, please mention that.  And typically I like to name the methods that have been added directly.

olivierlemasle: Thanks, I've updated the changelog.

dralley (approved):

PR #180, opened 2023-10-07
Add method to check existence of payload separated digest for package metadata

TommyLike:
Add the ability to check payload seperated digest for package metadata. see [rpm](https://github.com/rpm-software-management/rpm/blob/master/sign/rpmgensig.c#L601)

### 📜 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Support for setting file capabilities via the RPMTAGS_FILECAPS header.
 - `PackageMetadata::get_file_entries` method can get capability headers for each file.
+- Support check existence of package payload separated digest via `PackageMetadata::is_payload_separate_digest_present`
 
 ## 0.12.0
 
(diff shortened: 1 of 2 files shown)


dralley:
@TommyLike What is the goal of having that be a predefined method?  It seems very very niche.

TommyLike:
> rpm

Based on my understanding, external system can use this to detect whether to sign payload as well as header.

TommyLike:
> @TommyLike What is the goal of having that be a predefined method? It seems very very niche.

Based on my understanding, external system can use this to detect whether to sign payload as well as header.

dralley:
Why not put that logic in the external system itself, though?  Especially since v3 signatures will only be supported for a few more months.

TommyLike:
> Why not put that logic in the external system itself, though? Especially since v3 signatures will only be supported for a few more months.

yeah, this change will make it work for external system.

dralley:
I'm confused

TommyLike:
> I'm confused

Is there any other way that we can get the digest tag information for external system currently? I am not quite sure about current rpm APIs.

dralley:
I believe those methods (`package.header.entry_is_present(..)`) are public, so you can just call them in your own code.

PR #184, opened 2023-10-16
Bump MSRV, update dependencies

dralley:


olivierlemasle (commented):
@dralley Why 1.71.0?

Even if `time`s [_Minimum Rust version policy_](https://github.com/time-rs/time#minimum-rust-version-policy) **guarantees** that at least the three latest Rust versions are supported, its latest version (0.3.30) is currently supporting rust **>= 1.67.0**:

See `time`'s latest MSRV:

https://github.com/time-rs/time/blob/v0.3.30/time/Cargo.toml#L6

The other `rpm`'s dependencies and `rpm` itself also work with rust 1.67.0.

dralley:
We could use 1.67, but I would kinda rather do a bigger jump just to avoid needing to bump it again in another month, since this happens semi-frequently (unfortunately).

The main reason to keep an older MSRV is to make the package usable for building distro packages.  1.71 is currently available in CentOS 8 / 9 Stream, so I just went with that. 

Of course, the situation is not exactly ideal, I wish either `cargo` was smart enough to pick a suitable version of `time` or our dependencies would avoid so many version bumps.

dralley:
You're not wrong, it flows endlessly downhill.

OK.  I'll look at switching it to 1.67

dralley:
@olivierlemasle Is 1.67 OK?

olivierlemasle (approved):
Yes, thanks @dralley!

olivierlemasle (commented):
BTW, I've just noticed that three of de `dev-dependencies` are currently unused:
- `rsa`
- `rsa-der` (which depends on `simple_asn1`, bringing `time` in the dependency tree)
- `pretty_assertions`

When removing these three `dev-dependencies`, `cargo test --all-features` still work (and the MSRV could then be 1.66).

PR #185, opened 2023-10-17
Support signature with passphrase-protected PGP keys

### LINKED ISSUE #179: Support signature with password-protected PGP keys
The PGP signer works only if no the PGP private key is not protected by a passphrase.

Actually, an empty string is provided to PGP signer as the password:

https://github.com/rpm-rs/rpm/blob/bb5c1ea3b598ebd3e03969508057ab9e1d7ab359/src/rpm/signature/pgp.rs#L62-L64

The `rpm` crate should allow passing a password to PGP signer.

olivierlemasle:
Add method `with_key_passphrase` to `signature::pgp::Signer`, to provide the passphrase when the PGP secret key is passphrase-protected.

I'm to suggestions for a better API.

### 📜 Checklist

Closes #179


olivierlemasle:
CI failures are related to MSRV, and will be fixed with #184.

src/rpm/signature/pgp.rs
```diff
@@ -98,14 +99,24 @@ impl Signer {
...
             PublicKeyAlgorithm::EdDSA => Ok(Self {
                 secret_key,
                 algorithm: AlgorithmType::EdDSA,
+                key_passphrase: None,
             }),
             algorithm => Err(Error::UnsupportedPGPKeyType(algorithm)),
         }
     }
+
+    /// Returns a new [Signer], with the PGP key passphrase provided.
```
dralley, line 113: Technically it returns the same signer (while destroying the first copy via. move).  A better way of wording this might be.

`/// Configures the [Signer] with the provided PGP key passphrase`

dralley (commented):
This looks good!  Can you add a script with the commands to regenerate the password protected keys as was done here? 

https://github.com/rpm-rs/rpm/blob/master/test_assets/fixture_packages/configure_certs.sh  

Or perhaps move the keys and edit the existing scripts?  I'd like have the fixtures themselves be easy to extend and reproduce.

olivierlemasle:
@dralley I've moved the keys but I've created a separate script to re-generate them, because I was unsure it fitted in `configure_certs.sh`. Please tell me if you prefer otherwise.

dralley (approved):
This looks good, thanks!

olivierlemasle:
Thank you for the review!

PR #188, opened 2023-10-24
Unify file-related information under FileEntry (get_file_entries)

marxin:
Support all file-related functions for a case where an RPM is free of files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add method `is_no_replace` to `FileOptionsBuilder`, used to set the
   `%config(noreplace)` flag on a file.
 - Added the `FileEntry.linkto` field that is a target of a symbolic link.
+- Removed `get_file_checksums`, `get_file_ima_signatures` and `get_file_digest_algorithm`
+  function and the information is accessible via `FileEntry` struct
+  (`FileEntry::digest_string` and `FileEntry::ima_signature`).
+- Function `get_file_entries` returned empty vector for an RPM file without any file.
 
 ## 0.12.1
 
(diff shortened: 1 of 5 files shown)


dralley:
We have an "empty" rpm with no files [here](https://github.com/rpm-rs/rpm/tree/master/test_assets/fixture_packages), I'd prefer if you used that instead of adding a new one

dralley:
Also, I would kind of prefer to delete these methods entirely, and just make sure their functionality is present via `get_file_entries()`.  I don't see much point in getting a list of file checksums or signatures without having access to either the contents or at least the path of the file.

marxin:
> We have an "empty" rpm with no files [here](https://github.com/rpm-rs/rpm/tree/master/test_assets/fixture_packages), I'd prefer if you used that instead of adding a new one

Thanks, I've just done that.

marxin:
> I don't see much point in getting a list of file checksums or signatures without having access to either the contents or at least the path of the file, and making the user join them back together manually wouldn't be great API.

Yes, I also agree with the suggested approach. The only small drawback of the `get_file_entries` approach is that you need to parse many RPM Tag sections before you can assemble a `FileEntry`.

dralley:
>   Yes, I also agree with the suggested approach. The only small drawback of the get_file_entries approach is that you need to parse many RPM Tag sections before you can assemble a FileEntry.

What do you mean?  The whole header is "parsed" up-front.  It does clone a few values, and the timestamps and digests have a little bit of parsing, but I don't think it's overly expensive.

marxin:
> What do you mean? The whole header is "parsed" up-front.

Ah, ok, that's what I mean. So if it's parsed, then it's fine. If you want, I can prepare yet another pull request where I'll unify all under the umbrella of `get_file_entries` function.

marxin:
The question is if you want to do it instead of this PR, or after that. What do you prefer?

dralley:
You can use this PR

dralley:
note to self: I really ought to finish my own PR that will allow fetching the contents of file entries out of the archive

marxin:
Do you want to omit the `get_file_paths` function from the public API?

dralley:
That one I think could potentially be useful on its own (and isn't trivial to implement) so it should probably stay.

dralley (approved):
This looks great, thanks!

PR #189, opened 2023-10-24
Add FileEntry::linkto for symbolic links

marxin:


src/rpm/package.rs
```diff
@@ -843,8 +843,13 @@ impl PackageMetadata {
             Err(Error::TagNotFound(_)) => Ok(None),
             Err(e) => return Err(e),
         };
+        let links = self
+            .header
+            .get_entry_data_as_string_array(IndexTag::RPMTAG_FILELINKTOS);
```
dralley, line 848: Presuming an RPM has no files with symlinks whatsoever, is `RPMTAG_FILELINKTOS` guaranteed to exist?  If yes, this is fine, if no, then it should be treated like `RPMTAG_FILECAPS` a few lines above so that we don't erroneously return a bad result.

src/tests.rs
```diff
@@ -333,6 +333,18 @@ fn test_rpm_header() -> Result<(), Box<dyn std::error::Error>> {
 
     // Verify that if there are no capabilities set then the caps field is None
```
dralley, line 334: We might want to have a negative test for linktos similar to how this test is a negative test for capabilities.

src/rpm/package.rs
marxin: Don't know based on the documentation: https://rpm-software-management.github.io/rpm/manual/tags.html. However, I've just added a new symlink-free test-case and the tag is present. Hope it helps.

src/tests.rs
marxin: Yep. I've improved the assert macro usage where it expected an empty string for non-symlink entires and a non-empty one for a symlink.

src/rpm/package.rs
dralley: It looks like it adds the tag if there are files (regardless if any are symlinks), and always does if there are, and doesn't otherwise.  But in the event that there are no files then we should return no file entries, so it's fine.  

TL;DR this is OK, so long as there aren't RPMs floating around with different behavior.

dralley (approved):

PR #190, opened 2023-10-27
Fixup changelog and other small things

dralley:
@marxin

### DIFF
--- a/src/rpm/headers/header.rs
+++ b/src/rpm/headers/header.rs
@@ -383,56 +383,47 @@ pub struct FileOwnership {
 }
 
 #[derive(Debug, Clone, Hash, Eq, PartialEq)]
-pub enum FileDigest {
-    Md5(Vec<u8>),
-    Sha2_224(Vec<u8>),
-    Sha2_256(Vec<u8>),
-    Sha2_384(Vec<u8>),
-    Sha2_512(Vec<u8>),
-    // @todo unsupported other types for now
+pub struct FileDigest {
+    digest: String,
+    algo: DigestAlgorithm,
 }
 
 impl FileDigest {
-    pub fn load_from_str(
+    pub(crate) fn new(
         algorithm: DigestAlgorithm,
-        stringly_data: impl AsRef<str>,
+        hex_digest: impl Into<String>,
     ) -> Result<Self, Error> {
-        let hex: Vec<u8> = hex::decode(stringly_data.as_ref())?;
+        let hex = hex_digest.into();
+        let digest = FileDigest {
+            digest: hex,
+            algo: algorithm,
+        };
+
         Ok(match algorithm {
-            DigestAlgorithm::Md5 if hex.len() == 16 => FileDigest::Md5(hex),
-            DigestAlgorithm::Sha2_256 if hex.len() == 32 => FileDigest::Sha2_256(hex),
-            DigestAlgorithm::Sha2_224 if hex.len() == 30 => FileDigest::Sha2_224(hex),
-            DigestAlgorithm::Sha2_384 if hex.len() == 48 => FileDigest::Sha2_384(hex),
-            DigestAlgorithm::Sha2_512 if hex.len() == 64 => FileDigest::Sha2_512(hex),
+            DigestAlgorithm::Md5 if digest.digest.len() == 32 => digest,
+            DigestAlgorithm::Sha2_256 if digest.digest.len() == 64 => digest,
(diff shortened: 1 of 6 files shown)


src/rpm/package.rs
```diff
@@ -771,7 +771,7 @@ impl PackageMetadata {
     ///
     /// Note that this is not necessarily the same as the digest
     /// used for headers.
-    fn get_file_digest_algorithm(&self) -> Result<DigestAlgorithm, Error> {
+    pub fn get_file_digest_algorithm(&self) -> Result<DigestAlgorithm, Error> {
```
dralley, line 774: It occurred to me that actually, while the `FileDigest` is available, getting to the algorithm used would be a bit of a pain.  We should do a bit of work to ensure that both the string representation and the algorithm are trivially retrievable from `FileDigest`

marxin: Yep, you are right. It's unfortunate that `enum FileDigest` has the same variants as `enum DigestAlgorithm`. What about changing that to something like:

```rust
pub struct FileDigest {
    algorithm: DigestAlgorithm,
    digest: Vec<u8>,
}
```

where we can implement `to_string` method that will do `hex::encode(digest)` ?

dralley: That would work, though I think it might be better to just implement methods, so that the implementation of the struct doesn't matter.

marxin: I'm leaving the decision up to you, don't have any strong opinion about it.

dralley: I'll leave it. Something like e.g. rpmlint might want a rule that rejects md5 checksums.

marxin: I like the pull request!

dralley:
Anyone able to review this?  @drahnr?

src/rpm/headers/header.rs
```diff
@@ -378,31 +383,44 @@ pub struct FileOwnership {
...
 #[derive(Debug, Clone, Hash, Eq, PartialEq)]
-pub enum FileDigest {
-    Md5(Vec<u8>),
-    Sha2_256(Vec<u8>),
-    Sha2_384(Vec<u8>),
-    Sha2_512(Vec<u8>),
-    Sha2_224(Vec<u8>),
+pub struct FileDigest {
+    digest: String,
+    algo: DigestAlgorithm,
```
drahnr, line 388: :+1:

drahnr (changes requested):
I think the change to `FileDigest` is reasonable. The new public API requires documentation, the newly public API requires an improved documentation.

dralley:
>I am not convinced the conversions on style are more idiomatic or readable compared to their original content.

What are you referring to specifically?

tests/parsing.rs
```diff
@@ -89,9 +89,9 @@ fn test_rpm_file_signatures() -> Result<(), Box<dyn std::error::Error>> {
     assert_eq!(
         signatures,
         [
-            Some(String::from("0302041adfaa0e004630440220162785458f5d81d1393cc72afc642c86167c15891ea39213e28907b1c4e8dc6c02202fa86ad2f5e474d36c59300f736f52cb5ed24abb55759a71ec224184a7035a78")),
-            Some(String::from("0302041adfaa0e00483046022100bd940093777b75650980afb656507f2729a05c9b1bc9986993106de9f301a172022100b3384f6ba200a5a80647a0f0727c5b8f3ab01f74996a1550db605b44af3d10bf")),
-            Some(String::from("0302041adfaa0e00473045022068953626d7a5b65aa4b1f1e79a2223f2d3500ddcb3d75a7050477db0480a13e10221008637cefe8c570044e11ff95fa933c1454fd6aa8793bbf3e87edab2a2624df460")),
+            Some("0302041adfaa0e004630440220162785458f5d81d1393cc72afc642c86167c15891ea39213e28907b1c4e8dc6c02202fa86ad2f5e474d36c59300f736f52cb5ed24abb55759a71ec224184a7035a78".to_string()),
```
drahnr, line 92: This seems rather random

dralley: :shrug: personal preference.  If you want me to revert it I can revert it.

dralley: @drahnr Should I change this or can I merge it?  I don't mind reverting it, I'm just disputing the notion that the changes are necessarily random.

drahnr: Please revert.

dralley: Please re-review

drahnr (approved):

Issue #193, opened 2023-11-09
Support for building rpms on the fly


Lunarequest:
Currently I'm working on a tool that generates rpms from yaml part of the issue is that I will never know the files in the final rpm until the build is completed. Currently with `PackageBuilder` does not implement Copy which means it's impossible to generate a rpm on the fly using a loop. the old package rpm-rs while having a clunkier interface supported this. the two solutions i can see are

a) add the copy trait to `PackageBuilder`, this may cause issues we can not foresee.
b) implement a .with_files which takes a vec of tuples with (source, options)

dralley:
>the old package rpm-rs while having a clunkier interface supported this

Where was that the case?

Lunarequest:
> > the old package rpm-rs while having a clunkier interface supported this
> 
> Where was that the case? The interface is mostly unchanged

here is some code using the old RPMBuilder api, https://github.com/Lunarequest/sddm2rpm/blob/mistress/src/rpm_build.rs#L43-L55 it was possible to iterate over and add files to the rpm. which the new Packagebuild this isn't possible

dralley:
It's possible in exactly the same way though?  This code (copied from your example & slightly simplified) compiles and executes fine

```Rust
use rpm;
use std::env;
use std::path::Path;
use std::process::exit;
use walkdir;

fn main() {
    let mut pkg = rpm::PackageBuilder::new(
        "test",
        "0.0.1",
        "GPLv2",
        "noarch",
        "autogenrated sddm theme rpm",
    )
    .compression(rpm::CompressionType::Zstd)
    .requires(rpm::Dependency::any("sddm"));

    let info = "suse";
    if info.contains("suse") {
        pkg = pkg.requires(rpm::Dependency::any("kf5-plasma"))
            .requires(rpm::Dependency::any("libqt5-qtquickcontrols"))
            .requires(rpm::Dependency::any("libqt5-qtvirtualkeyboard"));
    } else if info.contains("Fedora Linux") {
        pkg = pkg.requires(rpm::Dependency::any("libKF5Plasma5"))
            .requires(rpm::Dependency::any("qt5-qtquickcontrols"))
            .requires(rpm::Dependency::any("desktop-backgrounds-compat"));
    } else {
        eprintln!("unsupported distro falling back to common package names\nNote: this may not be all dependencies");
    }

    for entry in walkdir::WalkDir::new("./src/")
        .into_iter()
        .filter_map(|e| e.ok())
    {
[...]

PR #194, opened 2023-11-09
add` PackageBuilder::with_files` too allows batch addition of files to a rpm

Lunarequest:
This pr address #193 and adds a `with_files` method which allows adding multiple files in batches to the rpm. There is a test based of `with_file` and documentation as well.

### 📜 Checklist

### DIFF
--- a/src/rpm/builder.rs
+++ b/src/rpm/builder.rs
@@ -306,6 +306,26 @@ impl PackageBuilder {
         Ok(self)
     }
 
+    pub fn with_files(
+        mut self,
+        files: Vec<(impl AsRef<Path>, impl Into<FileOptions>)>,
+    ) -> Result<Self, Error> {
+        for (source, options) in files {
+            let mut input = fs::File::open(source)?;
+            let mut content = Vec::new();
+            input.read_to_end(&mut content)?;
+            let mut options = options.into();
+            if options.inherit_permissions {
+                options.mode = (file_mode(&input)? as i32).into();
+            }
+
+            let modified_at = input.metadata()?.modified()?.try_into()?;
+
+            self.add_data(content, modified_at, options)?;
+        }
+        Ok(self)
+    }
+
     fn add_data(
         &mut self,
         content: Vec<u8>,
--- a/src/rpm/builder.rs
+++ b/src/rpm/builder.rs
@@ -306,6 +306,30 @@ impl PackageBuilder {
         Ok(self)
     }
 
+    /// Add multiple file to the package.
+    ///
+    /// ```
+    /// # fn foo() -> Result<(), Box<dyn std::error::Error>> {
+    /// let files =vec![(
+    ///         "./awesome-config.toml",
+    ///         rpm::FileOptions::new("/etc/awesome/config.toml").is_config(),
+    ///     ),    ///     // file mode is inherited from source file
+    ///     (
+    ///         "./awesome-bin",
+    ///         rpm::FileOptions::new("/usr/bin/awesome"),
+    ///     ),
+    ///     (
+    ///         "./awesome-config.toml",
+    ///         // you can set a custom mode, capabilities and custom user too
+    ///         rpm::FileOptions::new("/etc/awesome/second.toml").mode(0o100744).caps("cap_sys_admin=pe")?.user("hugo"),
+    ///     )];
+    /// let pkg = rpm::PackageBuilder::new("foo", "1.0.0", "Apache-2.0", "x86_64", "some baz package")
+    ///     .with_files(files)?
+    ///     .build()?;
+    /// # Ok(())
+    /// # }
+    /// ```
+
     pub fn with_files(
         mut self,
         files: Vec<(impl AsRef<Path>, impl Into<FileOptions>)>,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   of a `description`. A new method `PackageBuilder::description` can be used to
   set a detailed description for a package; if not set, the description defaults
   to the `summary`.
+- Add method `PackageBuilder::with_files` to allow addition of multiple files to rpm at one shot
 - Add method `with_key_passphrase` to `signature::pgp::Signer`, to provide the
   passphrase when the PGP secret key is passphrase-protected.
 - Add method `is_no_replace` to `FileOptionsBuilder`, used to set the
--- a/src/tests.rs
+++ b/src/tests.rs
@@ -84,7 +84,76 @@ However, it does nothing.",
             assert_eq!(f.mode, FileMode::from(0o120644));
         }
     });
+    Ok(())
+}
 
+#[test]
+fn test_rpm_batch_builder() -> Result<(), Box<dyn std::error::Error>> {
+    let mut buff = std::io::Cursor::new(Vec::<u8>::new());
+    let files = vec![
+        (
+            "Cargo.toml",
+            FileOptions::new("/etc/awesome/config.toml")
+                .is_config()
+                .is_no_replace(),
+        ),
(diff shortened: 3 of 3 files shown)


src/rpm/builder.rs
```diff
@@ -306,6 +306,50 @@ impl PackageBuilder {
...
+    /// let pkg = rpm::PackageBuilder::new("foo", "1.0.0", "Apache-2.0", "x86_64", "some baz package")
+    ///     .with_files(files)?
+    ///     .build()?;
+    /// # Ok(())
+    /// # }
+    /// ```
+
+    pub fn with_files(
+        mut self,
+        files: Vec<(impl AsRef<Path>, impl Into<FileOptions>)>,
```
drahnr, line 335: Could we make these explicit generics? It allows the user to specify the types in case of ambiguities.

drahnr: I am not a particular fan of tuples.

drahnr (commented):
A few small nits, generally looks good. Thank you for your contribution!

dralley:
I've changed my thinking a bit on the whole API for building RPMs.

The trouble is that if you look at building an RPM of even average complexity, the builder approach with pass-by-value just doesn't scale very well.  By that I mean mostly that it isn't readable.

```rust
 let built_rpm = PackageBuilder::new(
        "rpm-basic",
        "2.3.4",
        "MPL-2.0",
        "noarch",
        "A package for exercising basic features of RPM",
    )
    .epoch(1)
    .release("5")
    .description("This package attempts to exercise basic features of RPM packages.")
    .vendor("Los Pollos Hermanos")
    .url("http://www.savewalterwhite.com/")
    .vcs("https://github.com/rpm-rs/rpm")
    .group("Development/Tools")
    .packager("Walter White")
    .compression(CompressionType::None)
    .build_host("localhost")
    .source_date(1681068559)
    .provides(Dependency::any("/usr/bin/ls"))
    .provides(Dependency::any("aaronpaul"))
    .provides(Dependency::any("breaking(bad)"))
    .provides(Dependency::config("rpm-basic", "1:2.3.4-5.el9"))
    .provides(Dependency::eq("rpm-basic", "1:2.3.4-5.el9"))
    .provides(Dependency::eq("shock", "33"))
    .requires(Dependency::script_pre("/usr/sbin/ego"))
    .requires(Dependency::config("rpm-basic", "1:2.3.4-5.el9"))
    .requires(Dependency::greater_eq("methylamine", "1.0.0-1"))
    .requires(Dependency::less_eq("morality", "2"))
    .requires(Dependency::script_post("regret"))
    .conflicts(Dependency::greater("hank", "35"))
    .obsoletes(Dependency::less("gusfring", "32.1-0"))
    .obsoletes(Dependency::less("tucosalamanca", "444"))
    .supplements(Dependency::eq("comedy", "0:11.1-4"))
    .suggests(Dependency::any("chilipowder"))
    .enhances(Dependency::greater("purity", "9000"))
    .recommends(Dependency::any("SaulGoodman(CriminalLawyer)"))
    .recommends(Dependency::greater("huel", "9:11.0-0"))
    .with_file(
        "./tests/assets/SOURCES/example_config.toml",
        FileOptions::new("/etc/rpm-basic/example_config.toml").is_config(),
    )?
    .with_file(
        "./tests/assets/SOURCES/multiplication_tables.py",
        FileOptions::new("/usr/bin/rpm-basic"),
    )?
    .with_file(
        // I didn't bother filling all of these in yet, obviously it's not complete
        "",
        FileOptions::new("/usr/lib/rpm-basic").mode(FileMode::Dir { permissions: 0o644 }),
    )?
    .with_file(
        "",
        FileOptions::new("/usr/lib/rpm-basic/module").mode(FileMode::Dir { permissions: 0o755 }),
    )?
    .with_file(
        "./tests/assets/SOURCES/module/__init__.py",
        FileOptions::new("/usr/lib/rpm-basic/module/__init__.py"),
    )?
    .with_file(
        "./tests/assets/SOURCES/module/hello.py",
        FileOptions::new("/usr/lib/rpm-basic/module/hello.py"),
    )?
    .with_file(
        "",
        FileOptions::new("/usr/lib/rpm-basic/module").mode(FileMode::Dir { permissions: 0o755 }),
    )?
    .with_file(
        "",
        FileOptions::new("/usr/share/doc/rpm-basic").mode(FileMode::Regular { permissions: 0o644 }),
    )?
    .with_file(
        "",
        FileOptions::new("/usr/share/doc/rpm-basic/README").is_doc(),
    )?
    .with_file(
        "./tests/assets/SOURCES/example_data.xml",
        FileOptions::new("/usr/share/rpm-basic/example_data.xml"),
    )?
    .with_file(
        "",
        FileOptions::new("/var/log/rpm-basic/basic.log").is_ghost(),
    )?
    .with_file(
        "",
        FileOptions::new("/var/tmp/rpm-basic").mode(FileMode::Dir { permissions: 0o755 }),
    )?
    .add_changelog_entry(
        "Walter White <ww@savewalterwhite.com> - 3.3.3-3",
        "- I'm not in the meth business. I'm in the empire business.",
        1623672000,
    )
    .add_changelog_entry(
        "Gustavo Fring <gus@lospolloshermanos.com> - 2.2.2-2",
        "- Never Make The Same Mistake Twice.",
        1619352000,
    )
    .add_changelog_entry(
        "Mike Ehrmantraut <mike@lospolloshermanos.com> - 1.1.1-1",
        "- Just because you shot Jesse James, don't make you Jesse James.",
        1619352000,
    )
    .build()?;
```

Maybe it would be better if we passed by reference instead.  This would mean you have to assign ownership to a variable for the builder before proceeding with setting new fields, but it would make it a lot easier to split things up into sections and use loops like:

```rust
let mut builder = PackageBuilder::new(
        "rpm-basic",
        "2.3.4",
        "MPL-2.0",
        "noarch",
        "A package for exercising basic features of RPM",
    );

builder
    .epoch(1)
    .release("5")
    .description("This package attempts to exercise basic features of RPM packages.")
    .vendor("Los Pollos Hermanos")
    .url("http://www.savewalterwhite.com/")
    .vcs("https://github.com/rpm-rs/rpm")
    .group("Development/Tools")
    .packager("Walter White");

builder
    .provides(Dependency::any("/usr/bin/ls"))
    .provides(Dependency::any("aaronpaul"))
    .provides(Dependency::any("breaking(bad)"))
    .provides(Dependency::config("rpm-basic", "1:2.3.4-5.el9"))
    .provides(Dependency::eq("rpm-basic", "1:2.3.4-5.el9"))
    .provides(Dependency::eq("shock", "33"));

for req in requirements {
    builder.requires(req);
}
```

You could do this before using assignments (as noted in https://github.com/rpm-rs/rpm/issues/193#issuecomment-1807352560), but this way is cleaner and doesn't lean so hard on the compiler to remove a bunch of memcpys.

This is just an idea, though.  I'm curious what others think.

@Lunarequest @cat-in-136

Lunarequest:
I quite prefer the api you've proposed to the current one. The current one feels rather clunky and obtuse compared to the proposed one

cat-in-136:
Your suggested new design is more friendly for new users.

I'm glad the suggested code works with my existing coding style of splitting processes like `builder = builder.xxx(); builder = builder.yyy()`; as used in my project cargo-generate-rpm. I'm accepting your suggestion because it can be applied through simple find-and-replace is a good one.

> This would mean you have to assign ownership to a variable for the builder before proceeding with setting new fields

I consider it is possible to write the chained-style code without the need for a builder variable as like std::process::Command.

dralley:
>I consider it is possible to write the chained-style code without the need for a builder variable as like std::process::Command.

You're probably right, I was just thinking that if `.build()` consumed the builder then you'd definitely need to assign it to variable first, but actually I'm not sure there's necessarily particular reason to do that.

dralley:
Ok, I've filed https://github.com/rpm-rs/rpm/issues/220, we can continue this discussion there.

PR #198, opened 2023-12-03
Make Error Send + Sync (fixes #191)

### LINKED ISSUE #191: Support `Sync` trait for `rpm::Error`
I would like to use a return type equal to `anyhow::Result<RPMFile>`  for a multi-threaded code and I face the following issue:

```
        let header = rpm::PackageMetadata::parse(&mut buf_reader).context("RPM parsing failed")?;
```

```
error[E0599]: the method `context` exists for enum `Result<PackageMetadata, Error>`, but its trait bounds were not satisfied
   --> src/main.rs:140:67
    |
140 |         let header = rpm::PackageMetadata::parse(&mut buf_reader).context("RPM parsing failed")?;
    |                                                                   ^^^^^^^ method cannot be called on `Result<PackageMetadata, Error>` due to unsatisfied trait bounds
    |
   ::: /home/marxin/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/result.rs:502:1
    |
502 | pub enum Result<T, E> {
    | --------------------- doesn't satisfy `_: Context<PackageMetadata, Error>`
    |
   ::: /home/marxin/Programming/rpm-rs/src/errors.rs:9:1
    |
9   | pub enum Error {
    | --------------
    | |
    | doesn't satisfy `rpm::Error: Sync`
    | doesn't satisfy `rpm::Error: anyhow::context::ext::StdError`
    | doesn't satisfy `rpm::Error: std::marker::Send`
    |
    = note: the following trait bounds were not satisfied:
            `rpm::Error: anyhow::context::ext::StdError`
            which is required by `Result<PackageMetadata, rpm::Error>: anyhow::Context<PackageMetadata, rpm::Error>`
            `rpm::Error: std::marker::Send`
            which is required by `Result<PackageMetadata, rpm::Error>: anyhow::Context<PackageMetadata, rpm::Error>`
            `rpm::Error: Sync`
            which is required by `Result<PackageMetadata, rpm::Error>: anyhow::Context<PackageMetadata, rpm::Error>`
```

udoprog:
Uses a different approach than the ones in comments of #191. Most errors do not need to be boxed, and introducing two error variants instead of having one very generic one (with static details) ends up not being too bad.

### 📜 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- `Error` now implements `Send + Sync`.
+- `Error` now implements `Send + Sync` (therefore, `Result<Package, Error>` now implements `Send + Sync`).
 
 ## 0.13.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+
+- `Error` now implements `Send + Sync`.
+
 ## 0.13.0
 
 ### Breaking Changes
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -1,4 +1,4 @@
-use std::io;
+use std::{io, str::Utf8Error};
 
 use thiserror::Error;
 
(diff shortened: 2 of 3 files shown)


src/errors.rs
```diff
@@ -110,3 +122,8 @@ impl From<TimestampError> for Error {
         Error::TimestampConv(error)
     }
 }
+
+const _: () = {
+    const fn assert_send_sync<T: Send + Sync>() {}
+    assert_send_sync::<Error>();
+};
```
dralley, line 129: What is this?

udoprog: It asserts that the `Error` type implements `Send + Sync` at compile time.

dralley: Can we do this in a test rather than in the source code?  And leave documentation to that effect?

udoprog: This is just FYI, I don't particularly oppose moving it to a test if that's what you want to do.

Since these are auto traits, it could mean that the `Error` type loses its `Send + Sync`-ness due to an upstream change. You might want to keep this as an earlier, more appropriately placed compilation error to catch that.

dralley: Well, let's at least document what it is doing

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+
+- `Error` now implements `Send + Sync`.
```
dralley, line 12: Do you think it would be useful to extend this with

> (therefore, `Result<Package, Error>` now implements `Send + Sync`)

src/errors.rs
```diff
@@ -110,3 +122,8 @@ impl From<TimestampError> for Error {
         Error::TimestampConv(error)
     }
 }
+
```
udoprog, line 125:
```suggestion
// Assert at compile-time that Error implements Send and Sync.
```

CHANGELOG.md
udoprog, line 12:
```suggestion
- `Error` now implements `Send + Sync` (therefore, `Result<Package, Error>` now implements `Send + Sync`).
```

dralley (approved):

PR #200, opened 2023-12-05
feat. Adding options for %pretrans %posttrans scriptlets

### LINKED ISSUE #199: [Feature-Request] Adding `%pretrans` `%posttrans` scripts to builder
When rpm upgrades a package it will run all the install scripts of the new package before the uninstall scripts of the old package. 

The only hooks that are able to control overall pre/post steps seems to be %pretrans and %posttrans hooks of the new package.

Would it be possible to add these scripts to the Builder?

juliusl:
- During package upgrade, this gives the new package a hook after the old package is uninstalled.

Closes #199

### 📜 Checklist

### DIFF
--- a/src/rpm/builder.rs
+++ b/src/rpm/builder.rs
@@ -36,48 +36,6 @@ fn file_mode(_file: &fs::File) -> Result<u32, Error> {
     Ok(0)
 }
 
-/// This macro is for configuring scriptlet content w/ the additional `flags` and `prog tags,
-/// 
-/// **Error** If the scriptlet option is passed without passing the script content an error will be returned. 
-/// 
-macro_rules! prepare_scriptlet {
-    ($rcv:ident, $records:ident, $offset:ident, $(([$script:ident, $script_tag:path], [$flags:ident, $flags_tag:path], [$prog:ident, $prog_tag:path])),*) => {
-        $(
-            if let Some(script) = $rcv.$script {
-                $records.push(IndexEntry::new(
-                    $script_tag,
-                    $offset,
-                    IndexData::StringTag(script),
-                ));
-    
-                if let Some(flags) = $rcv.$flags {
-                    $records.push(IndexEntry::new(
-                        $flags_tag,
-                        $offset,
-                        IndexData::Int32(vec![flags]),
-                    ))
-                }
-    
-                if let Some(prog) = $rcv.$prog {
-                    $records.push(IndexEntry::new(
-                        $prog_tag,
-                        $offset,
-                        IndexData::StringArray(prog),
-                    ))
-                }
-            } else {
-                if $rcv.$flags.is_some() {
-                    return Err(Error::ScriptletOptionSetWithoutScript { option: stringify!($flags), scriptlet: stringify!($script) } );
-                }
-
-                if $rcv.$prog.is_some() {
-                    return Err(Error::ScriptletOptionSetWithoutScript { option: stringify!($prog), scriptlet: stringify!($script) } );
-                }
-            }
-        )*
-    };
-}
-
 /// Create an RPM file by specifying metadata and files using the builder pattern.
 #[derive(Default)]
 pub struct PackageBuilder {
@@ -106,37 +64,14 @@ pub struct PackageBuilder {
     enhances: Vec<Dependency>,
     supplements: Vec<Dependency>,
 
-    pre_inst_script: Option<String>,
-    pre_inst_flags: Option<u32>,
-    pre_inst_prog: Option<Vec<String>>,
-
-    post_inst_script: Option<String>,
-    post_inst_flags: Option<u32>,
-    post_inst_prog: Option<Vec<String>>,
-
-    pre_uninst_script: Option<String>,
-    pre_uninst_flags: Option<u32>,
-    pre_uninst_prog: Option<Vec<String>>,
-
-    post_uninst_script: Option<String>,
-    post_uninst_flags: Option<u32>,
-    post_uninst_prog: Option<Vec<String>>,
-
-    pre_trans_script: Option<String>,
-    pre_trans_flags: Option<u32>,
-    pre_trans_prog: Option<Vec<String>>,
-
-    post_trans_script: Option<String>,
-    post_trans_flags: Option<u32>,
-    post_trans_prog: Option<Vec<String>>,
-
-    pre_untrans_script: Option<String>,
-    pre_untrans_flags: Option<u32>,
-    pre_untrans_prog: Option<Vec<String>>,
-
-    post_untrans_script: Option<String>,
-    post_untrans_flags: Option<u32>,
-    post_untrans_prog: Option<Vec<String>>,
+    pre_inst_script: Option<Scriptlet>,
+    post_inst_script: Option<Scriptlet>,
+    pre_uninst_script: Option<Scriptlet>,
+    post_uninst_script: Option<Scriptlet>,
+    pre_trans_script: Option<Scriptlet>,
+    post_trans_script: Option<Scriptlet>,
+    pre_untrans_script: Option<Scriptlet>,
+    post_untrans_script: Option<Scriptlet>,
 
     /// The author name with email followed by a dash with the version
     /// `Max Mustermann <max@example.com> - 0.1-1`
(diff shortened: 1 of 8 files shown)


src/rpm/builder.rs
```diff
@@ -64,10 +64,12 @@ pub struct PackageBuilder {
     enhances: Vec<Dependency>,
     supplements: Vec<Dependency>,
 
+    pre_trans_script: Option<String>,
     pre_inst_script: Option<String>,
     post_inst_script: Option<String>,
     pre_uninst_script: Option<String>,
     post_uninst_script: Option<String>,
+    post_trans_script: Option<String>,
```
dralley, line 72: Can you put the two *_trans_script members together?

src/rpm/builder.rs
```diff
@@ -368,6 +370,11 @@ impl PackageBuilder {
         Ok(())
     }
 
+    pub fn pre_trans_script(mut self, content: impl Into<String>) -> Self {
+        self.pre_trans_script = Some(content.into());
+        self
+    }
```
dralley, line 376: Would you mind adding docstrings for all 6 methods here?

dralley: There's also some relevant information to know if you are using this feature.

This is a copy and paste, roughly, of what is already here for the other tags.  Unfortunately I'm pretty sure that what is already here is wrong or at least incomplete, including for the existing scriptlet tags.  I haven't had much time to look into it much yet but here's what I noticed:

For each of the scriptlet types there are actually 3 tags, each, there's the scriptlet itself, and also a set of flags and the command / arguments needed to execute it: https://github.com/rpm-software-management/rpm/blob/master/docs/manual/tags.md?plain=1#L189-L251

We're not currently doing anything with flags nor arguments.  I have no idea how important that is (as I said, haven't gotten around to it, and also I've never used these tags personally), but it is definitely incomplete.  There also appears to be `preuntrans` and `postuntrans` tags, which could probably be added in this PR.

IF you are up for it, it would be great if you could take a look into this and fix up the implementation a bit.  However, I don't want to pile on a bunch of extra work for you, and I won't require it to merge this PR or anything.  Any existing broken-ness is really a separate issue.

dralley:
Also needs a changelog entry

src/rpm/builder.rs
juliusl: @dralley Sure no problem, seems like it could be useful down the road. 

As an aside, do you happen to know why the scriptlet naming is different here, https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/#_syntax compared to here https://rpm-software-management.github.io/rpm/manual/tags.html?

Just want to double check my doc-/comments are accurate. I'm guessing rpm.org is probably the source of truth?

juliusl:
@dralley Addressed all comments and added the flags/prog tag code.

src/rpm/builder.rs
```diff
@@ -64,12 +106,37 @@ pub struct PackageBuilder {
     enhances: Vec<Dependency>,
     supplements: Vec<Dependency>,
 
-    pre_trans_script: Option<String>,
     pre_inst_script: Option<String>,
```
dralley, line 109: It might make sense to treat these like `Dependency` and create a common struct.

That is:

```
struct Scriptlet {
    script: Option<String>,
    flags: Option<u32>,
    prog: Option<Vec<String>>,
}
```

juliusl: @dralley Cool, but I'll leave the original builder methods for back-compat. Should I add a [deprecate] tag though?

dralley: There's no particular need to keep them for compat.  It'll just lead to more churn if we use a worse name and have to rename them back later.  Plus it ought to be a simple fix for the limited cases that were previously supported.

juliusl: @dralley I didn't happen to see this message before I pushed this most recent iteration, but let me know what you think. I think it's a reasonable middle-ground since setting flags/program is optional I optimized for setting scriptlets w/o the added config in the builder. For more advanced scenarios, I added a `scriptlet` method to the package builder.

src/rpm/package.rs
```diff
@@ -540,6 +548,64 @@ impl PackageMetadata {
...
     }
 
+    fn get_scriptlet(
+        &self,
+        scriptlet_tag: IndexTag,
+        flags_tag: IndexTag,
+        program_tag: IndexTag,
+    ) -> Result<Scriptlet, Error> {
+        let script = self.header.get_entry_data_as_string(scriptlet_tag);
+        let flags = self.header.get_entry_data_as_u32_array(flags_tag);
```
dralley, line 558: The documentation and code comments imply that this ought to be a lone integer rather than an integer array, although I know that some fields are "fuzzy" in this respect.  What is the motivation for treating it as an array of integers?

dralley:
Will finish looking at this tomorrow or Friday

src/rpm/package.rs
juliusl: @dralley oops, I was following along with Dependency's implementation and I noticed that it used u32_array for it's flags field and is also uses a u32 bitflag so I just thought that was the right option. I changed it to *_u32 and seems to still work.

src/rpm/builder.rs
```diff
@@ -368,26 +372,145 @@ impl PackageBuilder {
...
+                },
+                POSTUNTRANS_TAGS => {
+                    self.post_untrans_script = Some(scriptlet);
+                },
+                _ => {
+                    return Err(Error::NoScriptletTagsSet)
+                }
+            },
+            _ => {
+                return Err(Error::NoScriptletTagsSet)
```
dralley, line 507: This seems like a error that shouldn't need to exist?  Thinking in terms of a package, there cannot be an untyped scriptlet, so why allow one to be created?

src/tests.rs
```diff
@@ -51,6 +51,22 @@ However, it does nothing.",
                 .symlink("/usr/bin/awesome"),
         )?
         .pre_install_script("echo preinst")
+        .scriptlet(
```
dralley, line 54: I'm not sure we need two entirely different APIs to add a scriptlet.  The nice thing about an API like

`.pre_install_script(script: Into<Scriptlet>)`

 is that you could in theory do

`.pre_install_script("echo preinst")` if you have an `impl From<IntoString> for Scriptlet`

while still supporting more advanced things like

`.pre_install_script(
    Scriptlet::new("echo preinst")
        .flags(ScriptletFlags::EXPAND)
        .prog(vec!["/bin/blah/bash", "-c"])
)`

with the very same API.

Additionally it wouldn't seem like this meshes very well with the other functions?  For example, we don't do 

```
.dependency(Provides {...} )
.dependency(Requires {...} )
.dependency(Requires {...} )
.dependency(Requires {...} )
.dependency(Requires {...} )
```

and it seems like it would be less clear if we did use that kind of approach

juliusl: @dralley Thanks for taking the time to write this all out. It was honestly just a style choice I didn't have too strong of a reasoning behind it. I actually like your approach more.

src/tests.rs
```diff
@@ -84,6 +97,33 @@ However, it does nothing.",
             assert_eq!(f.mode, FileMode::from(0o120644));
         }
     });
+    
+    // Test scriptlet builder fn branches
+    let preinst = pkg.metadata.find_scriptlet(ScriptletType::PreInstall)?;
```
dralley, line 102: Could you split these in the same way as the setters, so that the pattern is the same?

src/rpm/headers/types.rs
```diff
@@ -382,6 +383,152 @@ impl<W: std::io::Write> std::io::Write for Sha256Writer<W> {
...
+    pub script: String,
+    /// Optional scriptlet flags,
+    ///
+    pub flags: Option<ScriptletFlags>,
+    /// Optional scriptlet interpreter/arguments,
+    ///
+    pub program: Option<Vec<String>>,
+    /// Type of scriptlet, (%prein, $postin, etc)
+    ///
+    pub ty: Option<ScriptletIndexTags>,
```
dralley, line 447: Once they're split, I'm not sure this is needed anymore?

src/rpm/headers/types.rs
```diff
@@ -382,6 +383,152 @@ impl<W: std::io::Write> std::io::Write for Sha256Writer<W> {
     }
 }
 
+/// Type-alias for a tuple containing index tags for a scriptlet type,
+///
+pub type ScriptletIndexTags = (IndexTag, IndexTag, IndexTag);
+
+/// Enumeration of different scriptlet types,
+/// 
+pub enum ScriptletType {
```
dralley, line 392: Nor this

src/rpm/headers/types.rs
```diff
@@ -505,4 +652,53 @@ mod test {
...
+        // Test full state
+        let _scriptlet = crate::Scriptlet::new(
+                r#"
+echo `hello world`
+        "#
+                .trim(),
+            )
+            .flags(crate::ScriptletFlags::EXPAND)
+            .prog(vec!["/usr/bin/blah", "-c"])
+            .ty(crate::ScriptletType::PreInstall);
```
dralley, line 667: Since we're passing it straight into `.pre_install_script()` or what have you, there's no real need to specify the type independently

dralley (approved):
Perfect, thanks!

Apologies for all the churn, but I really do appreciate the extra effort you put in to polish this up.  This is a solid improvement

It just needs a `cargo fmt` to clear up the CI errors

juliusl:
@dralley No worries, thanks for the review!

PR #218, opened 2024-03-02
Update dependencies & new release

dralley:
### 📜 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## 0.14.0
+
 ### Breaking Changes
 
+- Minimum supported Rust version updated to 1.71
 - `Dependency::rpmlib()` now inserts the `rpmlib()` portion automatically, only the feature name itself should
   be provided in the string passed as the name argument.
 - `FileOptions::is_no_replace()` is now `FileOptions::is_config_noreplace()` to reflect the fact that the noreplace
--- a/src/rpm/package.rs
+++ b/src/rpm/package.rs
@@ -178,17 +178,16 @@ impl Package {
         Ok(())
     }
 
-    /// Return the key id (issuer) of the signature as a hexadecimal string
+    /// Return the key ids (issuers) of the signature as a hexadecimal string
     #[cfg(feature = "signature-pgp")]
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -7,8 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## 0.14.0
+
 ### Breaking Changes
 
+- Minimum supported Rust version updated to 1.71
```
dralley, line 14: `env_logger` apparently bumped their MSRV and was causing issues when built on lower versions.

src/rpm/package.rs
```diff
@@ -178,40 +178,40 @@ impl Package {
         Ok(())
     }
 
-    /// Return the key id (issuer) of the signature as a hexadecimal string
+    /// Return the key ids (issuers) of the signature as a hexadecimal string
     #[cfg(feature = "signature-pgp")]
-    pub fn signature_key_id(&self) -> Result<Option<String>, Error> {
+    pub fn signature_key_ids(&self) -> Result<Vec<String>, Error> {
```
dralley, line 183: `rpgp` changed the underlying API here to reflect that signatures can apparently have more than one issuer / key ID.

CHANGELOG.md
drahnr: Since it's only  a dev dep, I am not sure we should bump Marv, but rather stick to `0.10.x`?

src/rpm/signature/pgp.rs
```diff
@@ -146,63 +146,8 @@ impl traits::Verifying for Verifier {
...
-                    }
-                } else {
-                    log::trace!(
-                        "Subkey key id {:?} does not match signature",
-                        sub_key.key_id()
-                    );
-                }
-            }
-            result
-        } else {
```
drahnr, line 205: I think some logic got removed, could you explain a bit why it's not needed?

drahnr: Never mind, it got moved down

drahnr (approved):
One question about msrv, otherwise LGTM

CHANGELOG.md
dralley: :shrug: We certainly could.  I figured a 7 months old version was still at least reasonable

PR #225, opened 2024-05-05
Delete v3 signature support for built packages

dralley:
RHEL and CentOS 7 are going out of support in a couple of weeks - obviously we won't want to cut a release until after that.

There's a few other simplifications that can be done that aren't part of this series currently, I'll probably do followup PRs for those.

### 📜 Checklist


dralley:
@cat-in-136 Are you OK with dropping EL7 support once it hits EOL in ~6 weeks?

cat-in-136:
I agree that it is time to retire support for legacy rpm version.

cmeister2:
I'll review the content later. Does this remove the ability to read and
write old packages, or just build them? Generally I'm loath to delete
functionality unless it's strictly necessary - there are companies where
they have to maintain old distro images for a long time and removing the
ability to interface with EL7 packages would mean they needed to fork the
crate.

If there is an undue maintenance cost then that's certainly worth
considering against the feature though.

dralley:
> I'll review the content later. Does this remove the ability to read and
write old packages, or just build them?

You should be able to read and write them fine.  Verifying signatures and digests for old packages would be a bit weaker because at the time there was no PAYLOADDIGEST in the header (added in 2015-2016 IIRC), so the only thing covering the payload integrity is the V3 signatures.

It does remove the ability to build them unless `[9df3d74](https://github.com/rpm-rs/rpm/pull/225/commits/9df3d7467012c212a31adeabd82d2d613ecfe038)`  is dropped from the series.

The primary maintenance benefit is that it allows us to generate the signature header from the primary header alone, rather than needing to look at both the primary header and the payload.  That simplifies the remote signing use case significantly (where you create a signature remotely from a checksum, rather than having access to the RPM).

drahnr:
Is there a use-case for verifying old packages still, while not generating any new packages with the legacy algorithms/features?

dralley:
You mean leaving in the code that verifies V3 signatures while dropping the code that generates them?

It's a reasonable half-step, generating them is the messiest part.  

Note that you can still verify the signatures of old packages with this patchset, and it's not completely meaningless (the header is signed, and the header has file digests), but yeah it would be neutered a bit.

drahnr:
> You mean leaving in the code that verifies V3 signatures while dropping the code that generates them?

Precisely. It wouldn't break verifying code, yet would stop producing broken signature schemes. I think a deprecation warning would also be in order.

To sum up, I propose a cadence of  three (3) releases.

1. immediate - add deprecation warnings on all legacy crypto types
2. weeks - delete the generation part of broken crypto schemes, keep the verficiation logic _with_ deprecation notices for obsolete crypto types
3. months - delete the verification logic, give people to migrate all their container workloads

Happy to hear your opinion

dralley:
Well in terms of "immediate", we don't have anything else to release at the moment... Dunno that's worth doing a release solely to announce deprecations

dralley:
@drahnr I've pared the PR down to merely removing the legacy signatures and digests from built, or resigned packages, and continuing to "verify" them.

Is that OK w/ you?

src/rpm/headers/header.rs
```diff
@@ -334,24 +334,34 @@ impl Header<IndexSignatureTag> {
         SignatureHeaderBuilder::<Empty>::new()
     }
 
+    /// The signature header is aligned to 8 bytes - the rest is filled up with zeroes.
+    ///
+    /// Parsing and writing out this section requires knowing how much padding is needed to complete
+    /// the alignment.
+    pub(crate) fn padding_required(&self) -> u32 {
+        (8 - (self.index_header.data_section_size % 8)) % 8
```
drahnr, line 342:
```suggestion
        8 - (self.index_header.data_section_size & 0b0111)
```

drahnr (approved):

src/rpm/headers/types.rs
```diff
@@ -277,6 +280,16 @@ impl FileOptionsBuilder {
         Ok(self)
     }
 
+    /// Direct which aspects of the file you would like RPM to verify.
+    ///
+    /// By default, every aspect of the file will be checked.
+    ///
+    /// See: `%verify` from specfile syntax
+    pub fn verify(mut self, flags: FileVerifyFlags) -> Self {
```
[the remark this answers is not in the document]
dralley: I can see how it would sound like an active operation in isolation, but since it's on the `FileOptions` type I think the meaning is probably sufficiently clear

src/rpm/headers/header.rs
```diff
@@ -339,7 +339,7 @@ impl Header<IndexSignatureTag> {
     /// Parsing and writing out this section requires knowing how much padding is needed to complete
     /// the alignment.
     pub(crate) fn padding_required(&self) -> u32 {
-        (8 - (self.index_header.data_section_size % 8)) % 8
+        8 - (self.index_header.data_section_size & 0b0111)
```
dralley, line 342: Is this from somewhere or did you do the math yourself?

drahnr: `0b0111` is masking down to the last bits, which is equiv to range `0..7`, which by definition is less than `8` and hence we omit both `modulo` ops.

dralley: You're absolutely 100% sure it's equivalent?  Because I once made a change that I thought was equivalent, and it was not, and the error looked a lot like the one now showing up in some of the integration tests. 

```
[stdout] /out/full_rpm_sig_protected.rpm:
[stdout] error: /out/full_rpm_sig_protected.rpm: hdr magic: BAD
```

src/rpm/headers/header.rs
```diff
@@ -334,24 +334,34 @@ impl Header<IndexSignatureTag> {
         SignatureHeaderBuilder::<Empty>::new()
     }
 
+    /// The signature header is aligned to 8 bytes - the rest is filled up with zeroes.
+    ///
+    /// Parsing and writing out this section requires knowing how much padding is needed to complete
+    /// the alignment.
+    pub(crate) fn padding_required(&self) -> u32 {
+        8 - (self.index_header.data_section_size & 0b0111)
```
drahnr, line 342:
```suggestion
        (8 - (self.index_header.data_section_size % 8) % 8)
```

src/rpm/headers/header.rs
dralley: Let's push this to a separate PR with unit tests if you want to continue with it

dralley: Looks like you made the same mistake as me btw.  Instead of `8 - (0...7)` it should be `8 - (1...8)`.  The first case leaves open the possibility of adding 8 bytes of padding which is not what you want.

drahnr: I did a revert, it's such a triviality of a stumbling block

PR #230, opened 2024-06-09
WIP: utilize binrw serialization/deserialization for Lead

marxin:
Hey. Recently, I've been playing with a great binary serialization and deserialization [library](https://binrw.rs/) and thought it would be cool to use it for the RPM project. What are the thoughts of the community about it?

I made a simple demonstration for the `Lead` struct, but I'm still struggling when it comes to more complex data like `IndexData` (where the content is split between The Index Entry and the actual data). I'm still working on that (https://github.com/jam1garner/binrw/discussions/267).

### DIFF
--- a/src/rpm/headers/lead.rs
+++ b/src/rpm/headers/lead.rs
@@ -1,6 +1,6 @@
-use nom::bytes::complete;
-use nom::number::complete::{be_u16, be_u8};
-use std::convert::TryInto;
+use binrw::io::NoSeek;
+use binrw::{BinRead, BinReaderExt, BinWrite, BinWriterExt};
+use std::io::Cursor;
 
 use crate::constants::*;
 use crate::errors::*;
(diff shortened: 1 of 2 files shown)


dralley:
To be completely honest I find this less straightforwards to read than the current code, which is pretty clean,

src/rpm/headers/lead.rs
```diff
@@ -10,119 +10,50 @@ use crate::errors::*;
...
     arch: u16,
-    name: [u8; 66],
+    #[br(map = |v: [u8; 66]| String::from_utf8_lossy(&v).to_string())]
+    #[bw(map = |name: &String| {
+        let mut name_arr = [0u8; 66];
+        let name_size = std::cmp::min(name_arr.len() - 1, name.len());
+        name_arr[..name_size].clone_from_slice(&name.as_bytes()[..name_size]);
+        name_arr
+    })]
+    name: String,
```
dralley, line 27: I would prefer to keep the data in the lead structure identical to what was on disk, rather than storing it as a potentially modified string

marxin:
> To be completely honest I find the macros less straightforwards to read than the current code, which is pretty clean and not particularly complex.

Sure, understood.

PR #231, opened 2024-06-24
Release 0.15.0

dralley:


.github/workflows/ci.yml
```diff
@@ -44,7 +44,7 @@ jobs:
       matrix:
         rust:
           - stable
-          - 1.71.0
+          - 1.72.0
```
dralley, line 47: Irritatingly, this actually has nothing to do with us, but because the `sha1` crate bumped their MSRV in a patch version.

PR #235, opened 2024-07-17
Speed up parse_header of the binary IndexData

marxin:
The change makes a measurable performance gain for debuginfod-rs:

```
Before:
  Time (mean ± σ):      2.322 s ±  0.042 s    [User: 25.047 s, System: 4.994 s]
After:
  Time (mean ± σ):      2.269 s ±  0.023 s    [User: 24.264 s, System: 5.174 s]

1.02 ± 0.02 times faster than debuginfod-rs-before
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+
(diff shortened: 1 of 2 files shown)


dralley (approved):
Please add a basic `### Changed` changelog entry stating that some optimizations were performed, no need to go into detail

marxin:
Good point, I've just added that.

PR #241, opened 2024-10-16
Add feature flags for compression algorithms

famfo:
This pull request adds a feature flag for every compression algorithm. All compression algorithms are all enabled by default, but can be disabled in case they are not used to reduce the amount of dependencies.

Please note: I couldn't run all the tests because some containers are not found in the registry I am using.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `zstdmt` feature which sets zstd compression to use all available cores.
 - Added feature flags for every compression algorithm to support disabling unused ones.
 
+### Breaking Changes
+
+- Changed default compression scheme from Gzip to Zstd.
+
 ## 0.15.0
 
 ### Breaking Changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - The `Header::parse_header` function gained a speed up related to parsing of the binary headers.
 - Added `zstdmt` feature which sets zstd compression to use all available cores.
+- Added feature flags for every compression algorithm to support disabling unused ones.
 
 ## 0.15.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Breaking Changes
 
 - Changed default compression scheme from Gzip to Zstd.
+- Removed bzip2 from the compression options enabled by default.
 
 ## 0.15.0
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,7 +41,7 @@ num = "0.4"
 enum-primitive-derive = "0.3"
 enum-display-derive = "0.1"
 cpio = "0.4"
-flate2 = "1"
+flate2 = { version = "1", optional = true }
 digest = "0.10"
 sha2 = "0.10"
 md-5 = "0.10"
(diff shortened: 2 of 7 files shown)


src/rpm/compressor.rs
```diff
@@ -107,37 +135,70 @@ impl Compressor {
...
+    #[cfg(feature = "gzip-compression")]
     Gzip(u32),
+    #[cfg(feature = "xz-compression")]
     Xz(u32),
+    #[cfg(feature = "bzip2-compression")]
     Bzip2(u32),
 }
 
 impl CompressionWithLevel {
+    // This function is used to check if a compression type is zstd in builder.rs:760. If zstd is
```
dralley, line 149: Line number is unstable, probably don't get that specific

src/rpm/compressor.rs
```diff
@@ -107,37 +135,70 @@ impl Compressor {
...
+            #[cfg(feature = "xz-compression")]
             Self::Xz(_) => CompressionType::Xz,
+            #[cfg(feature = "bzip2-compression")]
             Self::Bzip2(_) => CompressionType::Bzip2,
         }
     }
 }
 
 impl Default for CompressionWithLevel {
+    // The preference of the default compression type is listed in decending order and dependent on
```
dralley, line 168: I actually submitted a PR just now to default to Zstd.  Now that EL7 is end of life, we can do that without causing too many issues.

dralley: You may feel free to adopt that change and the changelog entry in this PR, for the sake of avoiding conflicts later.

famfo:
What should I do about the failing tests on `no-default-features`? I could disable the tests which need a specific compression for now.

Cargo.toml
```diff
@@ -63,12 +63,23 @@ gethostname = "0.5"
 hex-literal = "0.4"
 
 [features]
-default = ["signature-pgp"]
+default = [
+    "signature-pgp",
+    "gzip-compression",
+    "zstd-compression",
+    "xz-compression",
+    "bzip2-compression"
```
dralley, line 71: To be perfectly honest it might be reasonable to leave bzip2 out by default. It's relatively uncommon and isn't that great at compressing things in most circumstances compared to modern alternatives.

src/rpm/compressor.rs
```diff
@@ -107,37 +135,70 @@ impl Compressor {
...
+
+        #[cfg(feature = "zstd-compression")]
+        return CompressionType::Zstd.into();
+
+        #[cfg(feature = "xz-compression")]
+        return CompressionType::Xz.into();
+
+        #[cfg(feature = "bzip2-compression")]
+        return CompressionType::Bzip2.into();
+
```
dralley, line 186: I would not even have Bzip2 be a default option.

dralley:
>What should I do about the failing tests on no-default-features? I could disable the tests which need a specific compression for now.

Flag the test on the availability of the compression for now I suppose.

Could you explain your motivation for the PR - is it just reducing dependencies in a use case that doesn't require them?  (That's perfectly fine, I'm just curious)

famfo:
Mostly a use case that doesn't require anything else than zstd and we'd like to reduce the amount of C code we link against.

CHANGELOG.md
```diff
@@ -11,6 +11,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - The `Header::parse_header` function gained a speed up related to parsing of the binary headers.
 - Added `zstdmt` feature which sets zstd compression to use all available cores.
+- Added feature flags for every compression algorithm to support disabling unused ones.
+
+### Breaking Changes
+
+- Changed default compression scheme from Gzip to Zstd.
```
dralley, line 18: Add a note about moving Bzip2 to disabled by default.

src/rpm/compressor.rs
```diff
@@ -42,9 +54,11 @@ impl TryFrom<CompressionWithLevel> for Compressor {
     fn try_from(value: CompressionWithLevel) -> Result<Self, Self::Error> {
```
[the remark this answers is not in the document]
dralley: I'm open to counterargument though.  It's been a while since I thought about this and I don't actually use the library much myself currently.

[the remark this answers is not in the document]
famfo: Yup, you are right on that one, good catch.

dralley (approved):
Thanks.

dralley:
Do you have more changes planned?  If not I can do a new release, otherwise I can wait.

famfo:
Decompressing content from archives would still be nice but implementing that would probably still take a few days.

PR #243, opened 2024-11-07
tests: centos on docker hub went EOL

baloo:
centos docker image went EOL in June 2024 and do not appear to be available on hub.docker.com anymore.

This breaks tests when you don't already have an image cached on your computer.

This commit migrates to one available on quay.io

### 📜 Checklist

### DIFF
--- a/tests/compat.rs
+++ b/tests/compat.rs
@@ -22,8 +22,8 @@ mod pgp {
         let rpm_sig_check = format!("rpm -vv --checksig {} 2>&1;", path.as_ref().display());
         // TODO: check signatures on all distros?
         [
-            ("fedora:40", &rpm_sig_check),
-            ("fedora:40", &dnf_cmd),
+            ("quay.io/fedora/fedora:40", &rpm_sig_check),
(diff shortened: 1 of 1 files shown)


dralley:
CentOS Stream 9 is not EOL, it is an active distribution.

CentOS Stream is not the same as "CentOS".

dralley:
Reopening this because despite what I said being correct, you are also correct that the *Dockerhub* image is gone.

If you can get the tests to use quay.io instead of dockerhub, that would be preferable.

baloo:
> Reopening this because despite what I said being correct, you are also correct that the _Dockerhub_ image is gone.
> 
> If you can get the tests to use quay.io instead of dockerhub, that would be preferable. Likewise for Fedora.
> 
> https://quay.io/repository/centos/centos?tab=tags&tag=stream9

Ha nice, I didn't know there was one on quay.io

> Though that may be a system-default-settings issue rather than a test issue?

That could very well be, I'm not very knowledgeable about that.

dralley (approved):

PR #251, opened 2025-01-17
Add an API for reading files from the RPM payload

dralley:
### 📜 Checklist

Note: I'm working on tests, but they will be easier to merge as part of the test overhaul PR.  In the meantime I did some manual testing for this PR.

### DIFF
--- a/src/rpm/package.rs
+++ b/src/rpm/package.rs
@@ -1,5 +1,6 @@
 use std::{
     fs, io,
+    io::Read,
     path::{Path, PathBuf},
     str::FromStr,
 };
@@ -7,7 +8,7 @@ use std::{
 use digest::Digest;
 use num_traits::FromPrimitive;
 
-use crate::{constants::*, errors::*, CompressionType};
+use crate::{constants::*, decompress_stream, errors::*, CompressionType};
 
 #[cfg(feature = "signature-pgp")]
 use crate::signature::pgp::Verifier;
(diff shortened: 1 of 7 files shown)


dralley:
Here's a bit of example code for demo purposes:

```Rust
use rpm;
use std::env::args;

fn main() {
    let rpm_path = args().nth(1).expect("provide an RPM to extract");
    let package =
    rpm::Package::open(rpm_path).unwrap();

    let dest_path = args().nth(2).unwrap_or(package.metadata.get_name().unwrap().to_owned());

    package.extract(&dest_path).unwrap();

    for f in package.files().unwrap() {
        let f = f.unwrap();
        println!("{}", f.file_entry.path.display());
    }
}
```

src/rpm/package.rs
```diff
@@ -61,6 +63,94 @@ impl Package {
         self.write(&mut io::BufWriter::new(fs::File::create(path)?))
     }
 
+    /// Iterate over the file contents of the package payload
+    ///
+    /// # Examples
+    /// ```ignore
```
drahnr, line 69: There should be a space after any `# Titles`

src/rpm/package.rs
```diff
@@ -61,6 +63,94 @@ impl Package {
...
+
+        Ok(FileIterator {
+            file_entries,
+            archive,
+            count: 0,
+        })
+    }
+
+    /// Extract all contents of the package payload to a given directory
+    ///
```
drahnr, line 94: I think more detail is required on the behaviour, will it override existing files, absolute vs relative paths and security considerations.

drahnr (changes requested):
Looks good, bar some more documentation details

dralley:
@drahnr I tweaked the behavior slightly such that it uses `create_dir` instead of `create_dir_all`.  That simplifies the semantics.

src/rpm/package.rs
```diff
@@ -1019,3 +1137,53 @@ impl PackageMetadata {
...
+pub struct FileIterator<'a> {
+    file_entries: Vec<FileEntry>,
+    archive: Box<dyn io::Read + 'a>,
+    count: usize,
+}
+
+#[derive(Debug)]
+pub struct RpmFile {
+    pub metadata: FileEntry,
+    pub content: Vec<u8>,
```
drahnr, line 1150: Can we make this a slice or an offset regarding the actual package reader? I don't think we should read this to memory at all.

src/rpm/package.rs
```diff
@@ -61,6 +63,100 @@ impl Package {
...
+            count: 0,
+        })
+    }
+
+    /// Extract all contents of the package payload to a given directory.
+    ///
+    /// The if the directory is nested, its parent directories must already exist. If the
+    /// directory itself already exists, the operation will fail. All extracted files will be
+    /// dropped relative to the provided directory (it will not install any files).
+    ///
```
drahnr, line 99:
```suggestion
    /// # Implementation
    ///
    /// Currently the implementation reads each file to memory before writing it to disk
```

PR #260, opened 2025-03-08
Add signature clearing functionality

dralley:
### 📜 Checklist

### DIFF
--- a/src/rpm/package.rs
+++ b/src/rpm/package.rs
@@ -25,7 +25,7 @@ use super::Lead;
 ///
 /// Can either be created using the [`PackageBuilder`](crate::PackageBuilder)
 /// or used with [`parse`](`self::Package::parse`) to obtain from a file.
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 pub struct Package {
     /// Header and metadata structures.
     ///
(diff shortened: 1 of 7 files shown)


src/rpm/package.rs
```diff
@@ -159,6 +159,22 @@ impl Package {
...
+        let mut header_bytes = Vec::<u8>::with_capacity(1024);
+        // make sure to not hash any previous signatures in the header
+        self.metadata.header.write(&mut header_bytes)?;
+        let header_digest_sha256 = hex::encode(sha2::Sha256::digest(header_bytes.as_slice()));
+        let sig_header_builder =
+            Header::<IndexSignatureTag>::builder().add_digest(&header_digest_sha256);
+        self.metadata.signature = sig_header_builder.build();
+
+        Ok(())
+    }
```
dralley, line 176: This is the only true addition, the rest is mostly cleanups.

src/rpm/package.rs
```diff
@@ -159,6 +159,22 @@ impl Package {
         Ok(())
     }
 
+    // TODO: this should perhaps modify the header in-place instead?
```
dralley, line 162: This will wipe away all but the basic digest tags, which are really the most important ones (the rest being largely deprecated) but it's not the ideal implementation.  But clearing signatures in-place leaving any non-signature tags intact would be about an order of magnitude more complicated, so I'm not doing that right now.

dralley:
@drahnr

PR #262, opened 2025-03-08
Support "stripped cpio" format

### LINKED ISSUE #108: Drop "cpio" libraries and write something semi-custom, because RPM doesn't use vanilla CPIO

dralley:
closes #108 

### 📜 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added support for multiple signatures on a package as supported by upstream RPM.
 
+### Fixed
+
+- RPM packages that use large files (>4gb) now correctly declare rpmlib() dependency
+
 ## 0.17.0
 
 ### Added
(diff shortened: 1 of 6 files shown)


dralley:
As with the other PR, proper test coverage forthcoming in the testing overhaul

dralley:
I'll merge after a couple of days, until then I'll leave this up if you're able / willing to review.

PR #264, opened 2025-03-09
Multi signatures

dralley:
### 📜 Checklist


src/rpm/package.rs
```diff
@@ -242,37 +244,66 @@ impl Package {
...
+                decoder.read_to_end(&mut signature)?;
+                let signature = Verifier::parse_signature(&signature)?;
+
+                let new_key_ids: Vec<String> = signature
+                    .issuer()
+                    .iter()
+                    .map(|x| format!("{:x}", x))
+                    .collect();
+
+                // @TODO: is this right?  figure out what's up with multiple issuers per sig + multiple sigs
```
dralley, line 268: @DemiMarie Now that RPM packages can be signed by multiple keys, what is the most reasonable function signature for a function that returns the key IDs of the keys used to sign a package?

The APIs for fetching the signing key ID from a single signature already return `Vec<KeyId>`, presumably because it is theoretically possible to have multiple issuer subpackets (I am not familiar enough with OpenPGP to know if that is *reasonable* or *expected*, but the APIs seem to permit it).

I'm not sure if it is reasonable to flatten all of the `Vec<KeyId>` into a single `Vec<KeyId>` or if more structure should be provided?

dralley: Thanks.  Just curious, in what instance would you expect a signature to have more than one issuing key ID?

DemiMarie: I’m not aware of any, but I also don’t know enough about OpenPGP to know that there are not any.  In any case, either each `Vec` should be enforced to be of length 1 (with an error if they are not), or flattening them is a bad idea.

dralley: https://github.com/rpgp/rpgp/pull/262#discussion_r2019829648

DemiMarie: That answers if more than one issuer subpacket is possible, but not if no issuer subpackets can exist.

hko-s: The OpenPGP format allows for an impressive amount of flexibility, unfortunately.
Maybe it would be a good tradeoff for an somewhat opinionated application to model the issuer information as something like:

```
pub enum IssuerIdentity {
  Fingerprint(Fingerprint),
  KeyId(KeyId),
```

And when interpreting the Signature information, leaning towards preferring: Whatever subpackets are in the hashed area, if any, and the IssuerFingerprint, if both Issuer and IssuerFingerprint exist in the same (hashed or unhashed) area?

hko-s: (Also see my comment in https://github.com/rpgp/rpgp/pull/262#discussion_r2020047573 for more context)

src/rpm/package.rs
```diff
@@ -220,59 +222,90 @@ impl Package {
...
+                let new_key_ids: Vec<String> = signature
+                    .issuer()
+                    .iter()
+                    .map(|x| format!("{:x}", x))
+                    .collect();
+
+                if key_ids.len() != 1 {
+                    return Err(Error::UnexpectedIssuerCount(
+                        key_ids.len().try_into().unwrap(),
+                    ));
```
dralley, line 261: If there's more than one issuer packet, that's a clear error IMO.

If there are zero issuer packets, I'm OK with treating that as an error until I see a sufficient practical reason not to do so.

dralley:
@drahnr I'll merge after a couple of days, until then I'll leave this up if you have time to review.

src/rpm/package.rs
dralley: At least for now, until I find a practical reason to do otherwise, I'm tempted to just treat more or less than exactly 1 issuer packet as an error.  That keeps the API a simple flat `Vec<String>`.  

Given that this API has no users yet, and probably won't for the forseeable future, and will deal with a much smaller spectrum of PGP keys than `rpgp` in practice, I don't want to yak-shave too much :)

drahnr: I'd write an issue, and leave a breadcrumb reference for interested people to chime in in the issue.

dralley: Yeah at some point I need to go through all the TODO items in the codebase and make sure they have issues written up.  Or are dealt with.

src/rpm/package.rs
dralley: Not really, no

PR #266, opened 2025-03-21
Add Issuer as regular packet during signing.

### LINKED ISSUE #265: RPMs signed with rpm-rs are not valid with rpm in RHEL 7/8/9
I've created a small utility that fetches a PGP key from Hashicorp Vault and signs an RPM.  This works fine when tested with RPM on macos (from brew, 4.19.1.1), but when I try to verify the signature on RHEL servers there seem to an issue with the signature.  The Key is listed with ID of 00000000.

```
signed_rsa.rpm:
    Header V4 RSA/SHA256 Signature, key ID 00000000: NOKEY
    Header SHA256 digest: OK
    Payload SHA256 digest: OK
signed_dsa.rpm:
    Header V4 EdDSA/SHA256 Signature, key ID 00000000: NOKEY
    Header SHA256 digest: OK
    Payload SHA256 digest: OK
```

I've tried this with both RSA and DSA signatures.

On the macos rpm, the output looks fine.

```
signed_dsa.rpm:
    Header V4 EdDSA/SHA256 Signature, key ID 83eba43a: NOKEY
    Header SHA256 digest: OK
    Payload SHA256 digest: OK
signed_rsa.rpm:
    Header V4 RSA/SHA256 Signature, key ID d948f280: NOKEY
    Header SHA256 digest: OK
    Payload SHA256 digest: OK
```

I tried to dig into the rpm headers but ran out of time.

My guess is that the old PGP parser in RPM is failing while the newer rpm-sequoia does work.

I'm willing to spend some time digging into this issue, but might need some guidance on what to look for, and how best to diagnose the rpm headers.

thorhs:
RPM in RHEL 8 and 9 are unable to find the signing key id when signed with this the rpm crate.  Newer versions of RPM do not have this issue.

By adding a second Issuer packet which is not critical works around the old RPM deficiency.

Closes #265

### 📜 Checklist

### DIFF
--- a/src/rpm/signature/pgp.rs
+++ b/src/rpm/signature/pgp.rs
@@ -66,6 +66,11 @@ where
             .push(Subpacket::regular(SubpacketData::Issuer(
                 self.secret_key.key_id(),
             )));
+        sig_cfg
+            .hashed_subpackets
+            .push(Subpacket::regular(SubpacketData::IssuerFingerprint(
+                self.secret_key.fingerprint(),
+            )));
         //::pgp::packet::Subpacket::SignersUserID("rpm"), TODO this would be a nice addition
 
         let passwd_fn = || self.key_passphrase.clone().unwrap_or_default();
--- a/src/rpm/signature/pgp.rs
+++ b/src/rpm/signature/pgp.rs
@@ -60,10 +60,10 @@ where
         );
         sig_cfg
             .hashed_subpackets
-            .push(Subpacket::critical(SubpacketData::SignatureCreationTime(t)));
+            .push(Subpacket::regular(SubpacketData::SignatureCreationTime(t)));
         sig_cfg
             .hashed_subpackets
-            .push(Subpacket::critical(SubpacketData::Issuer(
+            .push(Subpacket::regular(SubpacketData::Issuer(
                 self.secret_key.key_id(),
             )));
         //::pgp::packet::Subpacket::SignersUserID("rpm"), TODO this would be a nice addition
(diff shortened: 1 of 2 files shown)


thorhs:
Just to be sure it does not get overlooked.  I am able to verify the signature of the RPM packages signed using the rpm crate on older versions of RPM.  The fix was to add the regular issuer packet, but I have no idea if it will cause issues for others reading RPM packages.

src/rpm/signature/pgp.rs
```diff
@@ -66,6 +66,11 @@ where
             .push(Subpacket::critical(SubpacketData::Issuer(
                 self.secret_key.key_id(),
             )));
+        sig_cfg
+            .hashed_subpackets
+            .push(Subpacket::regular(SubpacketData::Issuer(
+                self.secret_key.key_id(),
+            )));
```
dralley, line 73: Ideally we want to match what RPM is doing as closely as possible.  If standard RPM doesn't mark these subpackets critical and doesn't duplicate them (which based on the outputs you posted on the issue seems to be the case), then we probably shouldn't be doing that either.  

Could you try having just one "regular" copy of both the issuer and signature creation time?

Also, I think the issuer fingerprint can be handled with this:

```
     sig_cfg
            .hashed_subpackets
            .push(Subpacket::regular(SubpacketData::IssuerFingerprint(
                self.secret_key.fingerprint(),
            )));
```

thorhs: I've added the fingerprint as well, initially I wanted to do the minimum change to get signing to work.

thorhs:
@dralley I've refactored the commits a bit.  The first commit is refactoring how the test harness installs packages.  The differences between the distros were becoming troublesome so I moved it to the front of the test run and added some distro detection logic.

The second one adds the additional distros to test.

The third changes from critical to regular packets.

And the last one adds issuer fingerprint.

I decided to split them up for easier review, feel free to squash them.

dralley (approved):
Great work, thanks!

PR #269, opened 2025-04-24
Add methods to calculate metadata size

Richterrettich:
DNF repositories represent the binary RPM metadata in XML. Some metadata that is stored is related to sizes, e.G: `<rpm:header-range start="4504" end="17328"/>`
The given example metadata effectively says "Header starts at 4504 and ends at 17328"
This patch implements the necessary methods to calculate these sizes:
```rust
let end = pkg.metadata.get_total_metadata_size();
let start = end - pkg.metadata.header.size();
```

### 📜 Checklist

### DIFF
--- a/src/rpm/headers/header.rs
+++ b/src/rpm/headers/header.rs
@@ -284,7 +284,7 @@ where
     }
 
     /// Size (in bytes) of this header in on-disk representation, not including padding
-    pub(crate) fn size(&self) -> u32 {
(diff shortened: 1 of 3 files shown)


Richterrettich:
Ah nice, I've missed the offset method, thanks for pointing that out.  I'll take a look at your metadata project for sure. I think we can close this as the offset API already handles this.

dralley:
PRs welcome of course!

PR #271, opened 2025-05-27
chore(deps): bump `pgp` to `0.16`

baloo:
### 📜 Checklist


baloo:
This makes the upgrade to pgp 0.16 which introduces a couple API changes which affects rpm. There isn't a final release of pgp yet, but this brings the changes requires for the upgrade.

baloo:
This depends on https://github.com/rpgp/rpgp/pull/561 to fix clippy errors around size of the error enums.

src/rpm/headers/signatures.rs
```diff
@@ -61,7 +61,11 @@ impl SignatureHeaderBuilder {
             // need to base64-encode the raw bytes of the signatures
             for sig_bytes in &self.openpgp_signatures {
                 let signature = Verifier::parse_signature(sig_bytes)?;
-                let tag = match signature.config.pub_alg {
+                let tag = match signature
+                    .config()
```
dralley, line 65: Does the non-existence of the `config()` return value *actually* imply missing version or is it slightly different?

baloo: This is my understanding yes:
https://github.com/rpgp/rpgp/blob/80977cafbb01bdb675883a96e06a4e678cf8fa73/src/packet/signature/types.rs#L232-L237

https://github.com/rpgp/rpgp/pull/520

src/rpm/signature/pgp.rs
```diff
@@ -360,32 +362,38 @@ pub(crate) mod test {
...
             PublicKeyAlgorithm::RSA,
-            HashAlgorithm::SHA2_256,
+            HashAlgorithm::Sha256,
             [digest[0], digest[1]],
             signature,
             vec![
-                Subpacket::regular(SubpacketData::SignatureCreationTime(sig_time)),
-                Subpacket::regular(SubpacketData::Issuer(signing_key.key_id())),
+                Subpacket::regular(SubpacketData::SignatureCreationTime(sig_time)).unwrap(),
+                Subpacket::regular(SubpacketData::Issuer(signing_key.key_id())).unwrap(),
```
dralley, line 396: Can we use `expect()` here

src/rpm/signature/pgp.rs
```diff
@@ -415,60 +425,32 @@ pub(crate) mod test {
...
-        );
-        signer
-            .secret_key
-            .secret_subkeys
-            .iter()
-            .for_each(|public_subkey| {
-                log::debug!("secret subkeys in secret key: {:?}", public_subkey.key_id());
-                assert!(subkey_set.contains(public_subkey.key_id().as_ref()));
-            });
-    }
```
dralley, line 470: @baloo Do you remember why this test was removed

baloo: pgp changed the layout of its traits and I could not find a way to get that working.

PR #289, opened 2025-09-30
Add content length to signature header

Mark-Simulacrum:
This adds the RPMTAG_SIGSIZE/RPMTAG_LONGSIGSIZE to the signature header, which is required by some tooling to compute metadata about RPMs. Note that the actual constants used are from the signature header constants:

https://github.com/rpm-software-management/rpm/blob/eb4998108638df7588ce9cb7ee26de2f97da691d/include/rpm/rpmtag.h#L443

The non-signature constants cannot be put in the signature headers (the rpm tooling in the repo above throws an error when parsing such a file). So it seems likely this is the right way to encode the content length. It also avoids needing to update a header's value in place after writing it (so that the content length would account for the header indicating the length).

Documentation on these options:

* https://github.com/rpm-software-management/rpm/blob/master/docs/manual/tags.md#signatures-and-digests
* https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/8/html/packaging_and_distributing_software/new-features-in-rhel-8_packaging-and-distributing-software#ref_64-bit-rpm-tags_assembly_support-for-files-above-4-gb

The latter makes it explicit that this is the header + *compressed* payload.

### 📜 Checklist

   - Added tests that verify re-signing an archive produces the same size as the original (using the checked-in archives as baseline). This should confirm our computation is reasonably correct.
   - I'd expect it to, but unfortunately I can't easily run that locally (my distro is too old to have podman available). It looks like CI will test that for me though.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
-- Add support for payload length tag in the signature header. Newly built RPMs will have this set automatically.
+### Fixed
+
+- RPMv4 packages now populate payload length tags in the signature header. Newly built RPMs will have this set automatically.
+  If editing a pre-built RPM (e.g., for signing) the tag will be preserved if
+  already present. RPMv4 requires the presence of this tag.
 
 ## 0.18.2
 
--- a/src/rpm/builder.rs
+++ b/src/rpm/builder.rs
@@ -683,6 +683,7 @@ impl PackageBuilder {
 
     /// Build the package
     pub fn build(self) -> Result<Package, Error> {
+        let is_v4 = self.config.format == RpmFormat::V4;
         let (lead, header_idx_tag, content) = self.prepare_data()?;
 
         let mut header = Vec::with_capacity(128);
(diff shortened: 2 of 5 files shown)


dralley:
@Mark-Simulacrum Thanks for the report, but can I inquire a bit more about which software is looking at these tag values?

`RPMTAG_SIGSIZE` / `RPMTAG_LONGSIGSIZE` were intended for use with RPMv3 signatures, which have been legacy for more than 20 years at this point.  The latest RPM v6 spec (which is relatively recent to be fair) has removed those tags from the spec entirely. [0]

Likewise new versions of RPM don't create those signatures anymore unless you explicitly opt-in and therefore don't create these tags either.  rpm-rs doesn't have the ability to create those signatures anymore, we ripped that out a while back.  I'm open to adding the size tag back if there's a compelling compatibility reason but at the same time whatever software is looking at this tag is going to need to make some changes in relatively short order to deal with v6 packages, which will not have them.

[0] https://github.com/rpm-software-management/rpm/discussions/2919 - see "all size tags are dropped" under the "signature header" section

dralley:
>We currently are building packages intended to comply with the v4 format -- for Amazon Linux 2 -- which mandates the presence of these headers (and the native rpmbuild tooling appears to populate them)

Fair enough.

>Happy to gate the population of these tags on the v4 build config (https://docs.rs/rpm/latest/rpm/struct.BuildConfig.html#method.v4) perhaps?

That would be acceptable, thanks!

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+- Add support for payload length tag in the signature header. Newly built RPMs will have this set automatically.
```
dralley, line 10: Update the changelog entry also, and put it under a `### Fixed` or `### Changed` section heading. You can mention that this is aligning with RPMv4, which suggests that the tag should be present.

dralley:
>We currently are building packages intended to comply with the v4 format -- for Amazon Linux 2 

Neat!  Are you able to say which RPMs and whether it's `cargo-generate-rpm` or some specifically internal tool?  If not, no worries, I'm just curious.

Mark-Simulacrum:
> Neat! Are you able to say which RPMs and whether it's cargo-generate-rpm or some specifically internal tool? If not, no worries, I'm just curious.

We have an ad-hoc binary that uses this crate directly to build a package (we haven't looked at using cargo-generate-rpm, but are pretty happy with the interface this crate provides). My usage is for an internally-consumed RPM at this time.

Pushed a new commit with the changes (happy to squash or let you handle that during merge), let me know if there's any other changes to make.

Mark-Simulacrum:
Thank you for the quick response! Yes, I'm poking the relevant folks to try to chase down a migration path away from these headers.

Issue #252, opened 2025-01-17
It should be possible to read or extract data from the archive without loading the whole archive (or file?) into memory


dralley:
Currently `Package::open()` loads the entire archive into memory, while `PackageMetadata::open()` doesn't have access to the archive at all.

There's no way to stream access to the archive, or extract files from the archive, without loading the entire archive into memory and dumping each file (also extracted into a large buffer).  That's not ideal.

There should be some way to stream across the package and access files one at a time, preferably without needing to also fully dump  those into buffers also.

dralley:
Some progress has been made re: affiliated PRs.  Further ideas to try:

A) lazily read the package payload instead of pre-reading it into self.content, there could be an enum that either references a content buffer OR a file handle.

B) have FileIterator provide a reader handle (if possible) instead of a content buffer of individual file contents.  Though dropping the handle without reading might be tricky.  You'd have to avoid API users being able to mess up the state in various ways.

Not sure how realistic these are, but they would be nice if they could be made to work.

jamesarch:
I've submitted a PR addressing this: #345.

The implementation adds two public types under the existing `payload` feature:
- `PackageReader::open(path)` — reads only the package headers upfront; the payload is decompressed on demand via a `BufReader<File>` kept open.
- `StreamingRpmFile<'a>` — implements `Read` for one entry; unread bytes are drained automatically on drop so the next `next_file()` call is always correctly positioned.

Happy to adjust the API shape based on feedback.

PR #294, opened 2025-12-22
Reduce memory by copy file content to writer straight from the archive instead of…

syokensyo:
Reduce memory by copy file content straight from the archive to file, instead of reading each file entirely into memory before writing them when extract package. #252

### 📜 Checklist

### DIFF
--- a/src/rpm/package.rs
+++ b/src/rpm/package.rs
@@ -1,6 +1,6 @@
 use std::{
     fs,
-    io::{self, Read, Write},
+    io::{self, Read},
     path::{Path, PathBuf},
     str::FromStr,
 };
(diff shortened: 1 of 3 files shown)


src/rpm/package.rs
```diff
@@ -196,8 +200,8 @@ impl Package {
                     if file_path.exists() || file_path.symlink_metadata().is_ok() {
                         fs::remove_file(&file_path)?;
                     }
-
-                    symlink(file.metadata.linkto, &file_path)?;
+                    entry_reader.finish()?;
```
dralley: If `entry_reader.finish()` happens on every branch, is there a reason it can't be moved outside of the match block?

dralley (changes requested):

syokensyo:
> > Yes, it is better to do lazy read package payload, how about replace `content` with `package_path` in Package? but I'afraid this need some break changes, because we reuse Package in  `pub fn build(self) -> Result<Package, Error>` at build phase.
> 
> I was thinking it could be kind of like `Cow`, either an owned copy (buffer w/ `self.content` as it currently exists) or a reference to a package on the filesystem.

I think a reference to a package on the filesystem will be better.

> Probably I should just put these notes on the issue though. As I said I don't think they would need to block this PR

Yes, put notes on the issue will be more helpful.

PR #321, opened 2026-04-07
Make PackageBuilder API pass-by-reference

### LINKED ISSUE #220: Move from pass-by-value builder style API to pass-by-reference
The current style isn't very readable for RPMs of even average complexity, e.g.

```rust
 let built_rpm = PackageBuilder::new(
        "rpm-basic",
        "2.3.4",
        "MPL-2.0",
        "noarch",
        "A package for exercising basic features of RPM",
    )
    .epoch(1)
    .release("5")
    .description("This package attempts to exercise basic features of RPM packages.")
    .vendor("Los Pollos Hermanos")
    .url("http://www.savewalterwhite.com/")
    .vcs("https://github.com/rpm-rs/rpm")
    .group("Development/Tools")
    .packager("Walter White")
    .compression(CompressionType::None)
    .build_host("localhost")
    .source_date(1681068559)
    .provides(Dependency::any("/usr/bin/ls"))
    .provides(Dependency::any("aaronpaul"))
    .provides(Dependency::any("breaking(bad)"))
    .provides(Dependency::config("rpm-basic", "1:2.3.4-5.el9"))
    .provides(Dependency::eq("rpm-basic", "1:2.3.4-5.el9"))
    .provides(Dependency::eq("shock", "33"))
    .requires(Dependency::script_pre("/usr/sbin/ego"))
    .requires(Dependency::config("rpm-basic", "1:2.3.4-5.el9"))
    .requires(Dependency::greater_eq("methylamine", "1.0.0-1"))
    .requires(Dependency::less_eq("morality", "2"))
    .requires(Dependency::script_post("regret"))
    .conflicts(Dependency::greater("hank", "35"))
    .obsoletes(Dependency::less("gusfring", "32.1-0"))
    .obsoletes(Dependency::less("tucosalamanca", "444"))
    .supplements(Dependency::eq("comedy", "0:11.1-4"))
    .suggests(Dependency::any("chilipowder"))
    .enhances(Dependency::greater("purity", "9000"))
    .recommends(Dependency::any("SaulGoodman(CriminalLawyer)"))
    .recommends(Dependency::greater("huel", "9:11.0-0"))
    .with_file(
        "./tests/assets/SOURCES/example_config.toml",
        FileOptions::new("/etc/rpm-basic/example_config.toml").is_config(),
    )?
    .with_file(
        "./tests/assets/SOURCES/multiplication_tables.py",
        FileOptions::new("/usr/bin/rpm-basic"),
    )?
    .with_file(
        // I didn't bother filling all of these in yet, obviously it's not complete
        "",
        FileOptions::new("/usr/lib/rpm-basic").mode(FileMode::Dir { permissions: 0o644 }),
    )?
    .with_file(
        "",
        FileOptions::new("/usr/lib/rpm-basic/module").mode(FileMode::Dir { permissions: 0o755 }),
    )?
    .with_file(
        "./tests/assets/SOURCES/module/__init__.py",
        FileOptions::new("/usr/lib/rpm-basic/module/__init__.py"),
    )?
    .with_file(
        "./tests/assets/SOURCES/module/hello.py",
        FileOptions::new("/usr/lib/rpm-basic/module/hello.py"),
    )?
    .with_file(
        "",
        FileOptions::new("/usr/lib/rpm-basic/module").mode(FileMode::Dir { permissions: 0o755 }),
    )?
    .with_file(
        "",
        FileOptions::new("/usr/share/doc/rpm-basic").mode(FileMode::Regular { permissions: 0o644 }),
    )?
    .with_file(
        "",
        FileOptions::new("/usr/share/doc/rpm-basic/README").is_doc(),
    )?
    .with_file(
        "./tests/assets/SOURCES/example_data.xml",
        FileOptions::new("/usr/share/rpm-basic/example_data.xml"),
    )?
    .with_file(
        "",
        FileOptions::new("/var/log/rpm-basic/basic.log").is_ghost(),
    )?
    .with_file(
        "",
        FileOptions::new("/var/tmp/rpm-basic").mode(FileMode::Dir { permissions: 0o755 }),
    )?
    .add_changelog_entry(
        "Walter White <ww@savewalterwhite.com> - 3.3.3-3",
        "- I'm not in the meth business. I'm in the empire business.",
        1623672000,
    )
    .add_changelog_entry(
        "Gustavo Fring <gus@lospolloshermanos.com> - 2.2.2-2",
        "- Never Make The Same Mistake Twice.",
        1619352000,
    )
    .add_changelog_entry(
        "Mike Ehrmantraut <mike@lospolloshermanos.com> - 1.1.1-1",
        "- Just because you shot Jesse James, don't make you Jesse James.",
        1619352000,
    )
    .build()?;
```

You can do minor things like adding line breaks, but that's not the only issue.  Say you wanted to build an RPM from JSON, or user input or something at runtime, and you need to intertwine the builder with loops and if statements.  This can be done but you would need to constantly re-assign the variable e.g. https://github.com/rpm-rs/rpm/issues/193#issuecomment-1807352560

What we could potentially do instead, is switch to pass-by-reference style.  e.g.

```rust
let mut builder = PackageBuilder::new(
        "rpm-basic",
        "2.3.4",
        "MPL-2.0",
        "noarch",
        "A package for exercising basic features of RPM",
    );

builder
    .epoch(1)
    .release("5")
    .description("This package attempts to exercise basic features of RPM packages.")
    .vendor("Los Pollos Hermanos")
    .url("http://www.savewalterwhite.com/")
    .group("Development/Tools")
    .packager("Walter White");

if github_repo {
    builder.vcs("https://github.com/rpm-rs/rpm");
}

builder
    .provides(Dependency::any("/usr/bin/ls"))
    .provides(Dependency::any("aaronpaul"))
    .provides(Dependency::any("breaking(bad)"))
    .provides(Dependency::config("rpm-basic", "1:2.3.4-5.el9"))
    .provides(Dependency::eq("rpm-basic", "1:2.3.4-5.el9"))
    .provides(Dependency::eq("shock", "33"));

for req in requirements {
    builder.requires(req);
}
```

That would simplify intertwining logic into the builder construction.

dralley:
closes #220

### 📜 Checklist

### DIFF
--- a/src/rpm/builder.rs
+++ b/src/rpm/builder.rs
@@ -251,6 +251,9 @@ pub struct PackageBuilder {
     default_file_attrs: FileDefaults,
     /// Default ownership and permissions for directory entries (like the dirmode in `%defattr`).
     default_dir_attrs: FileDefaults,
+
+    /// Whether `build()` or `build_and_sign()` has already been called.
+    consumed: bool,
 }
 
 impl PackageBuilder {
@@ -295,7 +298,7 @@ impl PackageBuilder {
     /// # Ok(())
     /// # }
     /// ```
-    pub fn using_config(mut self, config: impl Into<BuildConfig>) -> Self {
+    pub fn using_config(&mut self, config: impl Into<BuildConfig>) -> &mut Self {
         self.config = config.into();
         self
     }
@@ -310,7 +313,7 @@ impl PackageBuilder {
     ///
     /// However, because of this, the epoch of a package must never decrease, and shouldn't be set
     /// unless required.
-    pub fn epoch(mut self, epoch: u32) -> Self {
+    pub fn epoch(&mut self, epoch: u32) -> &mut Self {
         self.epoch = Some(epoch);
         self
     }
@@ -331,45 +334,45 @@ impl PackageBuilder {
     /// * 3.fc38
     /// * 5.el9_2.alma
     /// * 0.20230715gitabcdef
-    pub fn release(mut self, release: impl Into<String>) -> Self {
+    pub fn release(&mut self, release: impl Into<String>) -> &mut Self {
         self.release = release.into();
         self
     }
 
     /// Set the URL for the package. Most often this is the website of the upstream project being
     /// packaged
-    pub fn url(mut self, content: impl Into<String>) -> Self {
+    pub fn url(&mut self, content: impl Into<String>) -> &mut Self {
         self.url = Some(content.into());
         self
     }
 
     /// Set the version control URL of the upstream project
-    pub fn vcs(mut self, content: impl Into<String>) -> Self {
+    pub fn vcs(&mut self, content: impl Into<String>) -> &mut Self {
         self.vcs = Some(content.into());
         self
     }
 
     /// Define a detailed, multiline, description of what the packaged software does
-    pub fn description(mut self, desc: impl Into<String>) -> Self {
+    pub fn description(&mut self, desc: impl Into<String>) -> &mut Self {
         self.desc = Some(desc.into());
         self
     }
 
     /// Set the package vendor - the name of the organization that is producing the package.
-    pub fn vendor(mut self, content: impl Into<String>) -> Self {
+    pub fn vendor(&mut self, content: impl Into<String>) -> &mut Self {
         self.vendor = Some(content.into());
         self
     }
 
     /// Set the packager, the name of the person producing the package. This is often not present,
     /// or set to the same value as the vendor
-    pub fn packager(mut self, content: impl Into<String>) -> Self {
+    pub fn packager(&mut self, content: impl Into<String>) -> &mut Self {
         self.packager = Some(content.into());
         self
     }
 
     /// Set the package group (this is deprecated in most packaging guidelines)
-    pub fn group(mut self, content: impl Into<String>) -> Self {
+    pub fn group(&mut self, content: impl Into<String>) -> &mut Self {
         self.group = Some(content.into());
         self
     }
@@ -386,7 +389,7 @@ impl PackageBuilder {
     /// # Ok(())
     /// # }
     /// ```
-    pub fn build_host(mut self, build_host: impl AsRef<str>) -> Self {
+    pub fn build_host(&mut self, build_host: impl AsRef<str>) -> &mut Self {
         self.build_host = Some(build_host.as_ref().to_owned());
         self
     }
@@ -395,7 +398,7 @@ impl PackageBuilder {
     /// one operation
     ///
     /// You can use any value, but the standard format is "${build_host} ${build_time}"
-    pub fn cookie(mut self, cookie: impl AsRef<str>) -> Self {
+    pub fn cookie(&mut self, cookie: impl AsRef<str>) -> &mut Self {
         self.cookie = Some(cookie.as_ref().to_owned());
         self
     }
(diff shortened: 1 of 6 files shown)


dralley:
I was considering making `summary` a method (`.summary()`) instead of an argument to `PackageBuilder::new()`.  It's a required field per rpmbuild though so it feels kinda borderline.

dralley:
Side question:

I want to add a Python API via PyO3.  Do you think it would be preferable to do that in a separate repo, or in this repo as a feature flag.

drahnr (changes requested):
I don't like the mut ref based changes, the ownership ensures a clear, compiler guaranteed flow which the reference based approach does not.

dralley:
>I don't like the mut ref based changes, the ownership ensures a clear, compiler guaranteed flow which the reference based approach does not.

I kinda get that, and tbh I was a bit underwhelmed by the ergonomic improvement (which wasn't nothing but also wasn't that much)

On the other hand, using `mut Self` copies something like 1200 bytes every time you call a method and doesn't necessarily get optimized out by the compiler (I think it's guaranteed not to actually, if it's in a loop).  `std::fs::OpenOptions` follows the same pattern, and I'm not sure that any other pattern would work for Python extensions.

If we're willing to pay the cost of an additional mass-clone, then we could leave the state intact after building a package, and that would resolve the re-use issue.  But IDK if that's a great idea either.

dralley:
The existing usage pattern still works FWIW - so long as you're not returning an incomplete builder early.   Very few test changes were needed.

src/rpm/builder.rs
```diff
@@ -237,6 +237,9 @@ pub struct PackageBuilder {
     default_file_attrs: FileDefaults,
     /// Default ownership and permissions for directory entries (like the dirmode in `%defattr`).
     default_dir_attrs: FileDefaults,
+
+    /// Whether `build()` or `build_and_sign()` has already been called.
+    consumed: bool,
```
drahnr, line 256: I don't think we should have this if we're using the reference based approach for `fn build*(..)` as well.

drahnr:
I agree that the struct is large to be copied around, on the other handside, a) copy is cheap and b) I don't quite see the use-case where the builder is run 1e6 times (or more). The `consumed` is a bit of a "wart" on the reference and I don't think the ergnomic trade-offs are quite there. If you have a compelling use-case to invalidate b) happy to toss my hypothesis and merge the PR.

dralley:
So the rationale for the overall change is basically outlined on the issue https://github.com/rpm-rs/rpm/issues/220, and I suppose you can add to that "pass by reference would make things much easier for a Python API".  I think without it a Python API would not be very usable if it could work at all.

In terms of `consumed`, I figured it was the least-bad option.  

Taking ownership would break the API harder (previous method calls take mutable references, so they cannot return ownership - so then in cases where previously you could still chain the whole thing, you'd have to assign the `PackageBuilder` to a variable always and call `bldr.build()` on a separate line.  So: that would work and maintain compile-time guarantees, but it would be a much more cumbersome API.

With a mutable reference, the other options without `consumed` are:

- Have `.build()` clone everything - even though the clone is very unlikely to be used
- Have `.build()` steal everything, and leave the builder in an unusable state, but without a proper error if you try to use it again.  I think it would fail on validating the empty (because it was stolen) name.

src/rpm/builder.rs
dralley: The other options were basically

* Have .build() clone everything
* Have .build() steal everything, and leave the builder in an unusable state, but without a proper error if you try to use it again

So I figured this was least-bad, since you could at least get a proper error.

dralley:
@drahnr So is any of that convincing or are you still partial to what exists currently.

In theory I could always avoid omit the `PackageBuilder` portions of the Python API and we can defer this decision a bit.

drahnr (approved):

dralley:
One other benefit to by-ref:  you can call methods that could fail and handle the errors instead of being instantly fatal to the `PackageBuilder` state.

PR #324, opened 2026-04-10
Remove obsolete v3 signature support, md5 checksums

dralley:
### 📜 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Removed
+
+- Removed verification of the MD5 header+payload digest (`RPMSIGTAG_MD5`) and the associated dependency on `md5`. MD5 is well and truly obsolete cryptography these days.
+- Removed verification of legacy "RPM v3"-style signatures.
+
 ## 0.20.0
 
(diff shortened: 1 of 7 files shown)


drahnr:
How common are rpm v3 signatures still?

I'd prefer to deprecate (on the generation side)  and/or hide them behind a feature flag (parsing side) if there is any reasonable usage still in the wild, with an explicit runtime opt-in on-top of the feature flag `fn allow_insecure_md5_digest(bool)` (or so) when parsing as a modifier.

dralley:
> I'd prefer to deprecate (on the generation side) and/or hide them behind a feature flag (parsing side) if there is any reasonable usage still in the wild, with an explicit runtime opt-in on-top of the feature flag fn allow_insecure_md5_digest(bool) (or so) when parsing as a modifier.

We actually phased them out on the generation side 2-ish years ago.  

https://github.com/rpm-rs/rpm/blob/master/CHANGELOG.md#changed-8

I think RPM hasn't included them by default for 5 or 6 years, OFC there it's still configurable and possible to add them

>How common are rpm v3 signatures still?

They're fairly often *present* but it's pretty rare for them to be the *only* signature present since RSAHEADER / DSAHEADER tags have been around since, like, 2001

The main utility of them is that before about 2015 the header didn't have a payload digest so the legacy signature was the only cryptographic tool that could attest to the integrity of the full RPM, header and payload.  Since around 2015 though there should always be a payload checksum in the main header covered by a header signature.

drahnr (approved):
Thank you for the clarification

PR #325, opened 2026-04-10
Detailed results on signature & digest verification

dralley:
### 📜 Checklist

### DIFF
--- a/src/rpm/package.rs
+++ b/src/rpm/package.rs
@@ -59,6 +59,165 @@ fn symlink(_original: &Path, _link: &Path) -> Result<(), Error> {
     Err(Error::UnsupportedSymlink)
 }
 
+/// The result of checking a single digest.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum DigestStatus {
+    /// The digest was present and matches the computed value.
+    Verified,
+    /// The digest tag was not present in the package headers.
+    NotPresent,
+    /// The digest was present but did not match the computed value.
+    Mismatch {
+        /// The value declared in the RPM header.
+        expected: String,
+        /// The value computed from the actual content.
+        actual: String,
+    },
+}
+
+impl DigestStatus {
+    /// Returns `true` if this digest was present and verified.
+    pub fn is_verified(&self) -> bool {
+        matches!(self, DigestStatus::Verified)
+    }
+
+    /// Returns `true` if this digest tag was not present.
+    pub fn is_not_present(&self) -> bool {
+        matches!(self, DigestStatus::NotPresent)
+    }
+
+    /// Returns `true` if this digest was present but mismatched.
+    pub fn is_mismatch(&self) -> bool {
+        matches!(self, DigestStatus::Mismatch { .. })
+    }
+
+    /// Check a digest by comparing the expected value (from a header lookup)
+    /// against the actual hash of the provided data.
+    ///
+    /// If the lookup returned an error (tag not present), returns `NotPresent`.
+    fn check_digest_against_tag<D: Digest>(expected: Result<&str, Error>, data: &[u8]) -> Self {
+        match expected {
+            Ok(expected) => {
+                let actual = hex::encode(D::digest(data));
+                if expected == actual {
+                    DigestStatus::Verified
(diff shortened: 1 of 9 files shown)


src/rpm/package.rs
```diff
@@ -59,6 +57,172 @@ fn symlink(_original: &Path, _link: &Path) -> Result<(), Error> {
...
+    }
+}
+
+/// Results of verifying all digests in the package.
+///
+/// Each field represents a specific digest type. Not all digests are present
+/// in all packages — v4 packages typically have MD5, SHA1, and SHA256, while
+/// v6 packages add SHA3-256 and SHA-512 variants.
+#[derive(Debug, Clone)]
+pub struct DigestReport {
```
drahnr, line 127: Shouldn't this be an `enum DigestReport { V4({SpecificV6, Common}), V6({SpecificV6, Common})}` with relevant accessors?

src/rpm/package.rs
```diff
@@ -59,6 +57,172 @@ fn symlink(_original: &Path, _link: &Path) -> Result<(), Error> {
...
+    pub payload_sha512: DigestStatus,
+    /// SHA3-256 of the compressed payload (v6 packages).
+    pub payload_sha3_256: DigestStatus,
+}
+
+impl DigestReport {
+    /// Collapse into a `Result`: fails if any digest mismatched.
+    ///
+    /// Digests that are [`DigestStatus::NotPresent`] are **not** considered failures.
+    pub fn result(&self) -> Result<(), Error> {
```
drahnr, line 162: I think the user should decide how to interpret the presence of `::NotPresent`.

src/rpm/package.rs
dralley: No - key word "typically" - different checksums got added at different times within the existence of "V4" and may or may not be present.  And you can add a V6 signature header to an otherwise V4 package with rpmsign.  On top of that, this library is currently capable of spitting out packages that use a superset of checksums for compatibility reasons, since there's no particular reason "why not" - so e.g. packages could have both SHA-1 and SHA3-256 checksums.

So I think trying to separate the digests into different sets would be overcomplicating things.

src/rpm/package.rs
dralley: Strictly speaking these are all public fields, so it's possible for a user to write their own function.  What would you want the API to look like otherwise?  You wouldn't expect all of these to be present basically ever.

It might make sense to assert that *at least one* checksum was verified for each type (payload / header) though.  Like, if *none* were present for some reason, that probably shouldn't succeed.

dralley: I pushed a commit with the described check, though honestly I think I would prefer to drop it apart from the `has_header_digest()` and `has_payload_digest()` methods.

I don't think want to be too prescriptive here honestly.  Imagine: users that might care about complying with certain standards for which sha1 or sha3 aren't "trusted".  We should either keep it simple, implement basic behavior and then leave the rest to the user or else go fully configurable - and I'm not prepared to do the latter right now.

src/rpm/package.rs
```diff
@@ -525,7 +541,19 @@ impl Package {
...
     pub fn verify_digests(&self) -> Result<(), Error> {
-        self.check_digests()?.result()
+        let digests = self.check_digests()?;
+
+        if !digests.has_header_digest() {
+            return Err(Error::NoHeaderDigestError);
+        }
+
+        // payload digests were added in 2017, so verification will fail on older RPMs
+        // for now, we just accept this - maybe we decide later that it needs to be more permissive
```
dralley: I feel like this might be a bit strong even now TBH.  SHA1HEADER has existed since 2001 so I have no qualms with assuming that at least one header digest will exist, but payload digests are another matter.

We do not actually claim to support RPMs prior to circa 2018 (roundabouts where RHEL8 was forked from) but it may still be a bit too much.

drahnr: We should probably warn for a bit, before making it failing 2018 feels somewhat too recent.

dralley: I removed the payloaddigest check and left a comment.

drahnr (approved):

PR #326, opened 2026-04-10
Create a Python API

dralley:
### 📜 Checklist


dralley:
@drahnr Do you have any interest in looking over any of this

If not it's ready to merge.

This should be the last major push from me in a good long while though.  I have some other projects to work on.

src/python.rs
```diff
@@ -0,0 +1,3418 @@
...
+                ("LICENSE", crate::FileFlags::LICENSE.bits()),
+                ("README", crate::FileFlags::README.bits()),
+                ("PUBKEY", crate::FileFlags::PUBKEY.bits()),
+                ("ARTIFACT", crate::FileFlags::ARTIFACT.bits()),
+            ],
+        )?,
+    )?;
+
+    m.add(
+        "DependencyFlags",
```
drahnr, line 3352: This is quite a lot of manual labor to maintain / upkeep

dralley: It shouldn't need to change much at this point - I'm pretty satisfied with the API now.  With all of the examples that are written now, any major usability issues are sorted out 

I really wish the constants list didn't need to be copied though.  I will admit that the cost/benefit of that is maybe not that high.  But it is at least just "mechanical".  The extension code is voluminous but not that complex.

dralley: I made a conscious effort to resolve every usability nitpick I could think of before doing this to avoid needing to duplicate work later.

PR #342, opened 2026-05-30
Optimizations: Make FileEntry borrowing

dralley:
### 📜 Checklist


src/rpm/content.rs
```diff
@@ -231,3 +239,1277 @@ impl ExactSizeIterator for FileIterator<'_> {
         self.file_entries.len() - self.count
     }
 }
+
+/// These tests cover payload + metadata integration, but they do equality tests on
+/// non-public fields / types, so they can't be actual integration tests, even though
+/// they otherwise would make more sense as integration tests.
```
dralley, line 245: In making the `FileEntry` fields private, which IMO is good from an architectural POV, it does create a bit of a dilemma.

The tests that were in `tests/payload.rs` check the integration of payload handling and header entry parsing.  The fields being tested are now private.  It is beneficial to look at those private details, because it A) helps visually document and verify that the internal implementation is correct, and B) ensures that nothing is missed later if a field were to be added or removed or something.

So, (I think) the most reasonable way to make those attributes visible to the tests is to move these tests into `src/` and duplicate any `common` code which was being used, which I have done.

I don't like this much but it felt like the best solution.

src/rpm/headers/header.rs
```diff
@@ -665,27 +673,125 @@ pub struct ChangelogEntry {
 
 /// User facing accessor type for a file entry with contextual information
 #[derive(Debug, Clone, Hash, Eq, PartialEq)]
-pub struct FileEntry {
-    /// Full path of the file entry and where it will be installed to.
-    pub path: PathBuf,
+pub struct FileEntry<'a> {
```
dralley, line 676: @drahnr The test changes inflate the diff a lot more than actually warranted, but the crux of the PR is this, would you like to weigh in?

Switching to methods for data access lets us change the underlying representation without breaking API, using `Cow` permits avoiding a ton of allocations while still making it own-able.

src/python.rs
```diff
@@ -924,19 +905,49 @@ impl PyPackageMetadata {
...
+            .map(|(dir, base)| {
+                let py_dir = dir_cache
+                    .entry(dir.as_ptr())
+                    .or_insert_with(|| pyo3::types::PyString::new(py, dir))
+                    .clone();
+                let py_base = pyo3::types::PyString::new(py, base);
+                Ok((py_dir, py_base))
+            })
+            .collect()
+    }
```
dralley, line 944: I have an actual use case for this, FWIW.  Re-using the same strings can save a shocking amount of memory at the scale of an entire repo.  I have code elsewhere that tries to accomplish the same thing from Python.

PR #345, opened 2026-06-15
Add PackageReader for streaming payload access without buffering

### LINKED ISSUE #252: It should be possible to read or extract data from the archive without loading the whole archive (or file?) into memory

jamesarch:
Closes #252.

## Problem

`Package::open()` reads the entire compressed payload into a `Vec<u8>` before returning. For callers that scan many packages (e.g. a createrepo-style tool building RPM repo metadata) or that only need a few files from a large package, this means unnecessary memory pressure — a 500 MB debuginfo RPM allocates 500 MB of RAM even if you only want one file.

## Solution

Two new public types, both inside the existing `payload` feature:

**`PackageReader`** — opens a package for streaming access:
```rust
let mut reader = rpm::PackageReader::open("my.rpm")?;
println!("{}", reader.metadata.get_nevra()?);

while let Some(mut file) = reader.next_file()? {
    println!("{}", file.metadata.path().display());
    // read file content on demand — no full-payload buffer
}
```

**`StreamingRpmFile<'a>`** — implements `Read` over one payload entry. Unread bytes are drained automatically via `payload::Reader::finish` when dropped, so the parent `PackageReader` is always left positioned at the start of the next entry.

## Design notes

- Ghost files are skipped transparently (they are not present in the payload archive).
- The existing `Package::open` / `Package::files()` / `Package::extract()` APIs are unchanged.
- `PackageReader` reuses the existing `payload::Reader<R>` machinery; the only new logic is keeping a `BufReader<File>` open instead of slurping into `Vec<u8>`.
- `StreamingRpmFile` holds `&'a mut Box<dyn Read>` into the parent's decompression stream, so the borrow checker enforces "finish one file before reading the next" at compile time.

## Tests

- `test_package_reader_matches_files_api`: streaming output matches `Package::files()` for v4 and v6 RPMs, across all non-ghost entries.
- `test_package_reader_partial_read_then_drop`: dropping a `StreamingRpmFile` mid-read (without consuming all bytes) leaves the reader correctly positioned.

```
cargo test --features payload --lib
test rpm::content::test_payload_integration::test_package_reader_matches_files_api ... ok
test rpm::content::test_payload_integration::test_package_reader_partial_read_then_drop ... ok
test result: ok. 61 passed; 0 failed
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+
+- `PackageReader` and `StreamingRpmFile`: streaming access to the payload archive without loading it into memory. `PackageReader::open()` reads only the package headers; payload bytes are decompressed on demand. `StreamingRpmFile` implements `Read` and provides an explicit `finish()` for propagating drain errors. Ghost files are returned with an empty body, matching `Package::files()`. Closes #252.
+
 ## 0.25.1
 
 ### Performance
--- a/src/rpm/content.rs
+++ b/src/rpm/content.rs
@@ -8,7 +8,7 @@ use std::os::unix::fs::PermissionsExt;
 use crate::{constants::*, decompress_stream, errors::*};
 
 use super::headers::*;
-use super::package::Package;
+use super::package::{Package, PackageMetadata};
 use super::payload;
 
 #[cfg(unix)]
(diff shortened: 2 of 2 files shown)


jamesarch:
Self-review pass before requesting maintainer attention — pushed an updated commit addressing the following:

**`#[cfg(feature = \"payload\")]` gate (was missing)**
Both `PackageReader` and `StreamingRpmFile`, along with all their `impl` blocks, are now gated. Verified with `cargo build --no-default-features`.

**Ghost file behaviour**
The original implementation silently skipped ghost files. `Package::files()` returns them with empty content, so `next_file()` now does the same — ghost entries are returned with `reader = None`, which makes `read()` return `Ok(0)` immediately. This keeps the two APIs consistent.

**Explicit `finish()`**
`Drop` drains remaining bytes best-effort and silences errors. Added `StreamingRpmFile::finish() -> io::Result<()>` so callers who care about truncation or IO errors can propagate them. `Drop` still calls it as a fallback.

**Doc comment**
Added a note explaining why `Iterator` is not implemented (lending-iterator constraint — `StreamingRpmFile` holds `&mut` into the decompression stream).

**CHANGELOG**
Added entry under `## Unreleased`.

All 62 lib tests pass with `--all-features`.

dralley:
The PR description isn't necessarily correct given that a `createrepo_c`-like tool can just use `PackageMetadata` and ignore the payload entirely, but overall this looks decent.  

I may tweak the API and changelog a bit post-merge.

dralley (approved):

PR #352, opened 2026-07-28
Add Package::decompress_payload() and ALT digest verification

dralley:
Add a method to decompress the payload in-place, leaving the raw (uncompressed) CPIO archive as the payload. The package metadata is not modified, so the PAYLOADCOMPRESSOR tag and primary digest tags still reflect the original compressed form.

Extend DigestReport with ALT payload digest fields (sha256_alt, sha512_alt, sha3_256_alt) that verify the uncompressed archive. DigestReport::result() now uses "either-set" semantics: the payload is valid if either the primary (compressed) or ALT (uncompressed) digest set has at least one verified entry with no mismatches. This allows verification to succeed both before and after decompression.

Assisted-By: Claude Opus 4.6

### 📜 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- `Package::decompress_payload()` decompresses the payload in-place, replacing compressed bytes with the raw archive.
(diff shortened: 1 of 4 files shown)


dralley:
@daandemeyer If you have any interest in reviewing this -- only part of your feature is relevant to this library as it doesn't install RPMs, but I'll implement the decompression (here) and payload alignment (future PR, if it gets accepted upstream) bits.

PR #353, opened 2026-07-30
Add optional Sequoia signature backend

### LINKED ISSUE #334: Sequoia-based PGP backend
The current backend is based on rpgp.  There's no reason to get rid of it, but it would potentially make sense to have a Sequoia-based backend selectable at build-time, since Sequoia is used by `rpm` upstream.

Minghan2005:
### Summary

- add an optional `signature-sequoia` backend using Sequoia 2.3 and its pure-Rust crypto implementation
- keep `signature-pgp` as the default while making the two signature backends mutually exclusive
- move signature metadata into backend-neutral types so either backend can inspect RPM signatures
- support armored keyrings, explicit key selection, protected keys, RPM v4/v6 signing and verification, and RSA, Ed25519, and ECDSA fixtures
- run the existing shared signature test suite against either backend and test both configurations independently in CI
- preserve the default rPGP draft-PQC support; Sequoia remains pinned to 2.3 because Cargo resolves optional dependencies into one lockfile, so Sequoia 2.4's stable `kem` API cannot coexist in this manifest with rPGP's draft-PQC pre-release `kem` API even when the features are mutually exclusive

### Validation

- `cargo test`
- `cargo test --no-default-features`
- `cargo test --no-default-features --features signature-pgp,payload,gzip-compression,zstd-compression,xz-compression,bzip2-compression,zstdmt`
- `cargo test --no-default-features --features signature-sequoia,payload,gzip-compression,zstd-compression,xz-compression,bzip2-compression,zstdmt`
- Clippy with default, minimal, and Sequoia configurations using `-D warnings`
- Rust 1.88 checks for both backend configurations
- `cargo fmt --all -- --check`

### Checklist

- [x] A changelog entry has been added
- [x] Documentation covers backend selection and the dependency constraint
- [x] Both backends use the shared signature suite
- [x] CI tests the mutually exclusive backend configurations in parallel

Closes #334

AI-assisted: Codex was used for implementation assistance. I reviewed and validated the change, and the commits include the required `Assisted-by` trailer.

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -68,9 +68,9 @@ pub enum Error {
     #[error("signature packet found, but no version was specified")]
     UnknownVersionSignature,
 
-    #[cfg(feature = "signature-pgp")]
+    #[cfg(any(feature = "signature-pgp", feature = "signature-sequoia"))]
     #[error("error creating signature: {0}")]
-    SignError(#[source] pgp::errors::Error),
+    SignError(#[source] Box<dyn std::error::Error + Send + Sync>),
 
     #[error("error parsing keys, failed to parse bytes as utf8 for ascii armored parsing")]
     KeyLoadUtf8Error(
@@ -79,23 +79,19 @@ pub enum Error {
         Utf8Error,
     ),
 
-    #[cfg(feature = "signature-pgp")]
+    #[cfg(any(feature = "signature-pgp", feature = "signature-sequoia"))]
     #[error("errors parsing keys, failed to parse bytes as ascii armored key")]
-    KeyLoadSecretKeyError(
-        #[from]
-        #[source]
-        pgp::errors::Error,
-    ),
+    KeyLoadSecretKeyError(#[source] Box<dyn std::error::Error + Send + Sync>),
 
     #[cfg(feature = "signature-pgp")]
     #[error("key binding signature verification failed: {0}")]
     KeyBindingVerificationError(pgp::errors::Error),
 
-    #[cfg(feature = "signature-pgp")]
+    #[cfg(any(feature = "signature-pgp", feature = "signature-sequoia"))]
     #[error("error verifying signature with key {key_ref}: {source}")]
     VerificationError {
         #[source]
-        source: pgp::errors::Error,
+        source: Box<dyn std::error::Error + Send + Sync>,
         key_ref: String,
     },
 
--- a/src/rpm/headers/signatures.rs
+++ b/src/rpm/headers/signatures.rs
@@ -2,19 +2,17 @@
 
 use super::*;
 
-#[cfg(feature = "signature-pgp")]
+#[cfg(feature = "signature-meta")]
 use std::collections::HashSet;
 use std::default::Default;
 
 use crate::RpmFormat;
 use crate::constants::*;
-#[cfg(feature = "signature-pgp")]
-use crate::signature::pgp::Verifier;
+#[cfg(feature = "signature-meta")]
+use crate::signature::{SignatureAlgorithm, parse_signature_info};
 
 use base64::prelude::*;
 use digest::Digest;
-#[cfg(feature = "signature-pgp")]
-use pgp::crypto::public_key::PublicKeyAlgorithm;
 
 /// base signature header builder
 pub struct SignatureHeaderBuilder {
@@ -43,7 +41,7 @@ pub(crate) fn decode_sig(signature: &str) -> Result<Vec<u8>, crate::Error> {
         .map_err(|e| crate::Error::Io(std::io::Error::new(std::io::ErrorKind::InvalidData, e)))
 }
 
-#[cfg(feature = "signature-pgp")]
+#[cfg(feature = "signature-meta")]
 pub(crate) fn encode_sig(signature: &[u8]) -> String {
     BASE64_STANDARD.encode(signature)
 }
@@ -157,7 +155,7 @@ impl SignatureHeaderBuilder {
     pub fn build(self) -> Result<Header<IndexSignatureTag>, crate::Error> {
         let mut entries = Vec::new();
 
-        #[cfg(feature = "signature-pgp")]
+        #[cfg(feature = "signature-meta")]
         if !self.openpgp_signatures.is_empty() {
             let mut openpgp_signatures = Vec::new();
             let mut legacy_sig = None;
@@ -167,25 +165,24 @@ impl SignatureHeaderBuilder {
             // precedence when deduplicating by issuer fingerprint, and the
             // legacy signature tag is set from the newest matching signature.
             for sig_bytes in self.openpgp_signatures.iter().rev() {
-                let signature = Verifier::parse_signature(sig_bytes)?;
+                let info = parse_signature_info(sig_bytes)?;
 
                 // Deduplicate: keep only the newest signature per key
-                if let Some(fp) = signature.issuer_fingerprint().first()
-                    && !seen_fingerprints.insert(format!("{fp:x}"))
+                if let Some(fp) = info.fingerprint()
+                    && !seen_fingerprints.insert(fp.to_owned())
                 {
                     continue;
                 }
 
                 if legacy_sig.is_none() {
-                    let legacy_sig_tag = match signature
-                        .config()
-                        .ok_or(crate::Error::UnknownVersionSignature)?
-                        .pub_alg
-                    {
-                        PublicKeyAlgorithm::RSA => Some(IndexSignatureTag::RPMSIGTAG_RSA),
-                        PublicKeyAlgorithm::ECDSA
-                        | PublicKeyAlgorithm::EdDSALegacy
-                        | PublicKeyAlgorithm::Ed25519 => Some(IndexSignatureTag::RPMSIGTAG_DSA),
+                    let legacy_sig_tag = match info.algorithm() {
+                        Some(SignatureAlgorithm::RSA) => Some(IndexSignatureTag::RPMSIGTAG_RSA),
+                        Some(
+                            SignatureAlgorithm::ECDSA
+                            | SignatureAlgorithm::EdDSALegacy
+                            | SignatureAlgorithm::Ed25519
+                            | SignatureAlgorithm::Ed448,
+                        ) => Some(IndexSignatureTag::RPMSIGTAG_DSA),
                         _ => None,
                     };
                     if let Some(legacy_sig_tag) = legacy_sig_tag {
--- a/src/rpm/package.rs
+++ b/src/rpm/package.rs
@@ -11,8 +11,6 @@ use num_traits::FromPrimitive;
 
 use crate::{CompressionType, constants::*, errors::*};
 
-#[cfg(feature = "signature-pgp")]
-use crate::signature::pgp::Verifier;
 #[cfg(feature = "signature-meta")]
 use crate::{Timestamp, signature};
 #[cfg(feature = "signature-meta")]
@@ -202,16 +200,16 @@ impl DigestReport {
 }
 
 /// The result of verifying a single signature against the provided keys.
-#[cfg(feature = "signature-pgp")]
+#[cfg(feature = "signature-meta")]
 #[derive(Debug)]
 pub struct SignatureCheckResult {
     /// Parsed metadata about the signature (fingerprint, algorithm, etc.).
-    pub info: signature::pgp::SignatureInfo,
+    pub info: signature::SignatureInfo,
     /// `None` if verified successfully, `Some(error)` if verification failed.
     pub error: Option<Error>,
 }
 
-#[cfg(feature = "signature-pgp")]
+#[cfg(feature = "signature-meta")]
 impl SignatureCheckResult {
     /// Returns `Ok(())` if verified, or `Err` with the verification error.
     pub fn result(&self) -> Result<(), &Error> {
@@ -228,7 +226,7 @@ impl SignatureCheckResult {
 }
 
 /// Results of verifying all digests and signatures in the package.
-#[cfg(feature = "signature-pgp")]
+#[cfg(feature = "signature-meta")]
 #[derive(Debug)]
 pub struct SignatureReport {
     /// Digest verification results.
@@ -237,7 +235,7 @@ pub struct SignatureReport {
     pub signatures: Vec<SignatureCheckResult>,
 }
 
-#[cfg(feature = "signature-pgp")]
+#[cfg(feature = "signature-meta")]
 impl SignatureReport {
     /// Collapse into a `Result`: fails if any digest mismatched or if no signature
     /// was successfully verified.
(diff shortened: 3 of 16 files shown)


Minghan2005:
Thank you for confirming the implementation direction in #334. The pull request workflow is currently marked action_required because this is a first-time contribution. Could a maintainer approve the CI run when convenient? The full all-features suite, independent feature configurations, Clippy, rustdoc, and the Rust 1.88 MSRV checks pass locally.

src/rpm/headers/signatures.rs
```diff
@@ -2,19 +2,17 @@
 
 use super::*;
 
-#[cfg(feature = "signature-pgp")]
+#[cfg(any(feature = "signature-pgp", feature = "signature-sequoia"))]
```
dralley: Use `signature-meta`, which ought to be implied by either `signature-pgp` or `signature-sequoia`.

src/rpm/headers/signatures.rs
```diff
@@ -2,19 +2,17 @@
...
 use std::collections::HashSet;
 use std::default::Default;
 
 use crate::RpmFormat;
 use crate::constants::*;
-#[cfg(feature = "signature-pgp")]
-use crate::signature::pgp::Verifier;
+#[cfg(any(feature = "signature-pgp", feature = "signature-sequoia"))]
```
dralley: Same

src/rpm/headers/signatures.rs
```diff
@@ -43,7 +41,7 @@ pub(crate) fn decode_sig(signature: &str) -> Result<Vec<u8>, crate::Error> {
         .map_err(|e| crate::Error::Io(std::io::Error::new(std::io::ErrorKind::InvalidData, e)))
 }
 
-#[cfg(feature = "signature-pgp")]
+#[cfg(any(feature = "signature-pgp", feature = "signature-sequoia"))]
```
dralley: Same all the way down

src/rpm/signature/meta.rs
```diff
@@ -0,0 +1,115 @@
...
+    fingerprint: Option<String>,
+    key_id: Option<String>,
+    version: SignatureVersion,
+    algorithm: Option<SignatureAlgorithm>,
+    hash_algorithm: Option<SignatureHashAlgorithm>,
+    created: Option<u32>,
+}
+
+impl SignatureInfo {
+    #[cfg(any(feature = "signature-pgp", feature = "signature-sequoia"))]
```
dralley: Compilation of this whole module ought to be implied by `signature-meta` so I don't think this would be necessary

src/rpm/signature/mod.rs
```diff
@@ -1,9 +1,29 @@
...
 pub mod pgp;
 
+#[cfg(feature = "signature-sequoia")]
+pub mod sequoia;
+
+#[cfg(feature = "signature-pgp")]
+pub(crate) fn parse_signature_info(signature: &[u8]) -> Result<SignatureInfo, crate::Error> {
+    Ok(pgp::signature_info(&pgp::Verifier::parse_signature(
+        signature,
+    )?))
```
dralley: Why use ? to unwrap and then wrap it in `Ok()`?  Could probably just return the Result directly.

src/rpm/package.rs
```diff
@@ -202,16 +200,16 @@ impl DigestReport {
 }
 
 /// The result of verifying a single signature against the provided keys.
-#[cfg(feature = "signature-pgp")]
+#[cfg(any(feature = "signature-pgp", feature = "signature-sequoia"))]
```
dralley: Same with all of these, use `signature-meta` instead

src/rpm/package.rs
```diff
@@ -647,7 +645,7 @@ impl Package {
     ///
     /// This is a convenience wrapper around [`check_signatures`](Self::check_signatures)
     /// that collapses the detailed report into a pass/fail `Result`.
-    #[cfg(feature = "signature-meta")]
```
dralley, line 657: Why change this?

src/rpm/signature/sequoia.rs
```diff
@@ -0,0 +1,400 @@
+//! OpenPGP signing and verification backed by Sequoia.
+//!
+//! This backend covers the v4 and v6 RSA, Ed25519, and ECDSA algorithms
+//! supported by Sequoia 2.3's pure-Rust crypto implementation.  The draft-PQC
```
dralley: Sequoia 2.4 is out, and it ought to support those algorithms now

https://gitlab.com/sequoia-pgp/sequoia/-/blob/main/openpgp/NEWS?ref_type=heads#L16

src/errors.rs
```diff
@@ -72,6 +72,18 @@ pub enum Error {
     #[error("error creating signature: {0}")]
     SignError(#[source] pgp::errors::Error),
 
+    #[cfg(feature = "signature-sequoia")]
+    #[error("error creating Sequoia signature: {0}")]
+    SequoiaSignError(String),
```
dralley: These errors and their rpgp equivalents are all conditionally compiled, so IMO we ought to go either in the direction of naming these the same (e.g. `SignError`) with a concrete instance of the underlying library's error whichever way it was compiled, or having one version of the error (named generically) with a string type inside.  Or a `Box<dyn Error>`.

src/rpm/headers/signatures.rs
Minghan2005: Updated to gate this import on `signature-meta` in `df5f8be`.

src/rpm/headers/signatures.rs
Minghan2005: Updated to use `signature-meta` here as well in `df5f8be`.

src/rpm/headers/signatures.rs
Minghan2005: Updated the remaining signature-header gates to `signature-meta` in `df5f8be`.

src/rpm/signature/meta.rs
Minghan2005: Agreed. `SignatureInfo::new` now inherits the module-level `signature-meta` gate without an extra backend condition.

src/rpm/signature/mod.rs
Minghan2005: Simplified this to return the mapped parser `Result` directly in `df5f8be`.

src/rpm/package.rs
Minghan2005: Updated the backend-neutral package types and methods to use `signature-meta` consistently.

src/rpm/package.rs
Minghan2005: This change was unnecessary. I restored the original `signature-meta` gate in `df5f8be`.

tests/sequoia_signatures.rs
```diff
@@ -0,0 +1,234 @@
+use std::fs;
```
dralley: One test single test module that tests whichever implementation is compiled in would be preferable.  It doesn't look like there's much sequoia-specific stuff here, so that could be isolated to functions or conditionally compiled blocks in the existing signature test file.

dralley: I suppose I am assuming that only one module can be enabled at a time

src/errors.rs
Minghan2005: Addressed in 09f9e9a. The rPGP and Sequoia paths now share the same string-backed `SignError`, `KeyLoadSecretKeyError`, and `VerificationError` variants, which also works when both backends are enabled together. This follows the backend-neutral string option and avoids exposing feature-dependent concrete error types. The full all-features suite, each backend independently, Clippy, and Rust 1.88 checks pass.

src/rpm/signature/sequoia.rs
Minghan2005: I tested Sequoia 2.4.1 directly. It does add ML-DSA support, but enabling it alongside rPGP 0.20 draft-PQC currently makes the dependency graph unbuildable: Sequoia pulls `ml-kem 0.3.2`/stable `kem 0.3`, while rPGP pulls `ml-kem 0.2.x` against `kem 0.3.0-pre.0`; Cargo resolves the latter to stable `kem 0.3.0`, then `ml-kem 0.2.1` fails with 18 API errors. Pinning the prerelease fixes rPGP but cannot satisfy Sequoia’s `kem = ^0.3`. Commit 09f9e9a clarifies this in the module docs. I have left this thread unresolved: would you prefer keeping Sequoia 2.3 until the rPGP PQC dependency converges, or dropping/decomposing rPGP draft-PQC support to permit Sequoia 2.4 now?

src/errors.rs
```diff
@@ -68,9 +68,9 @@ pub enum Error {
     #[error("signature packet found, but no version was specified")]
     UnknownVersionSignature,
 
-    #[cfg(feature = "signature-pgp")]
+    #[cfg(any(feature = "signature-pgp", feature = "signature-sequoia"))]
```
dralley, line 71: `signature-meta`

src/errors.rs
```diff
@@ -68,9 +68,9 @@ pub enum Error {
...
     #[error("error creating signature: {0}")]
-    SignError(#[source] pgp::errors::Error),
+    SignError(String),
```
dralley: Is there a reason to convert to string immediately instead of `Box<dyn Error>`?  There could be, I would just like to understand what it is, if so.

src/errors.rs
dralley: I don't know that there is a need to have both backends compiled at the same time.  If there is a good reason to do so, let me know.

But making it so that only one backend was enabled at a time would solve the conflicts between the two dependency chains (I think?) and also allow sharing one test suite.  The CI could of course test both on parallel builds.

src/rpm/signature/sequoia.rs
dralley: I don't know that there is a need to have both backends compiled at the same time. If there is a good reason to do so, let me know.

But making it so that only one backend was enabled at a time would solve the conflicts between the two dependency chains (I think?) and also allow sharing one test suite. The CI could of course test both on parallel builds.

Minghan2005: Implemented the mutually exclusive backend model in c4ab8fb: enabling both features now produces a compile error, and CI builds/tests rPGP and Sequoia as separate matrix entries. I also retried Sequoia 2.4 under that model, but Cargo still resolves all optional dependencies into one lockfile before feature selection. Its stable kem ^0.3 and rPGP draft-PQC kem 0.3.0-pre.0 are therefore forced to one version and either Sequoia or rPGP fails to compile. Preserving the existing default rPGP PQC support requires retaining Sequoia 2.3 for now; the manifest and module docs record that constraint.

tests/sequoia_signatures.rs
Minghan2005: Addressed in c4ab8fb. The separate Sequoia integration test target is removed; tests/signatures.rs now imports whichever backend feature is selected and runs the shared signing, verification, keyring, tampering, in-place signing, and roundtrip suite. Only draft-PQC cases that Sequoia 2.3 cannot support are conditionally limited to rPGP. CI runs this same test target once per backend.

src/errors.rs
Minghan2005: Addressed in df5f8be: the backend-neutral error variants are gated by signature-meta.

src/errors.rs
Minghan2005: I kept the shared errors string-backed in 09f9e9a so the public Error enum does not expose a feature-dependent boxed source type and both implementations produce the same stable variants. c4ab8fb now makes the implementations mutually exclusive, but this representation still keeps downstream matching and diagnostics backend-neutral.

dralley: A `Box<dyn Error>` would be a stable type between the two though, yes?  It's a form of type erasure.  What is the issue with it, specifically?

tests/signatures.rs
```diff
@@ -2,13 +2,25 @@ use std::fs;
...
 
+#[cfg(feature = "signature-pgp")]
+fn test_timestamp() -> rpm::Timestamp {
+    rpm::Timestamp(1_600_000_000)
+}
+
+#[cfg(feature = "signature-sequoia")]
+fn test_timestamp() -> rpm::Timestamp {
+    rpm::Timestamp::now()
+}
```
dralley, line 16: Can you explain why the timestamp used is different  for the sequoia backend tests?  Is that necessary for some reason?

dralley:
If you can use `Box<dyn Error>` instead of `String`, please do.  I'm fine with doing so for both error variants, not behind a feature flag.  It's still strictly better than `String`.

dralley:
Also, go ahead and squash the commits.

tests/signatures.rs
Minghan2005: The old `1_600_000_000` timestamp predates the test keys, which were created in March 2026. Sequoia correctly rejects a signature whose creation time precedes the signing subkey binding, while the rPGP path did not expose that policy check. I initially used `Timestamp::now()` to get past that validity boundary, but that made the test nondeterministic. Commit 3287b63 now uses the same fixed timestamp (`1_780_000_000`, 2026-05-28) for both backends, after the test keys were created; the full signature suite passes with each backend.

Minghan2005:
Addressed both follow-ups in the squashed commit 3287b63. The backend-neutral signing, key-loading, and verification variants now retain boxed error sources (`Box<dyn Error + Send + Sync>`) instead of converting them to strings, and the branch is now a single commit. I also replaced the Sequoia-only current timestamp with one deterministic timestamp shared by both backend test configurations. Both complete backend test suites, both signature integration suites, doctests, formatting, no-feature compilation, and the CI-equivalent Clippy checks pass locally.

Minghan2005:
The final squashed commit `3287b63` created a fresh first-time-contributor workflow run, which is currently `action_required`. All review threads are now addressed and resolved, and I re-ran the default, minimal, rPGP, and Sequoia test configurations plus strict Clippy and Rust 1.88 checks locally. When convenient, could a maintainer approve this final CI run? Thank you.

src/rpm/signature/sequoia.rs
```diff
@@ -0,0 +1,438 @@
+//! OpenPGP signing and verification backed by Sequoia.
+//!
+//! This backend uses Sequoia's pure-Rust crypto implementation. It is selected
+//! instead of, rather than alongside, the `signature-pgp` backend. Sequoia 2.3
+//! is used because Cargo resolves optional dependencies into one lockfile, and
+//! Sequoia 2.4's stable KEM API cannot be resolved alongside rPGP's draft-PQC
+//! pre-release KEM API even though the crate features are mutually exclusive.
```
dralley: Drop everything starting from "Sequoia 2.3 is used because" - it's unnecessary details for a code comment and likely to get out of date.

dralley:
Please also rebase on master branch, and fix up the changelog if it gets applied into the wrong section (should be under `## Unreleased`

src/rpm/signature/sequoia.rs
Minghan2005: Addressed in ec21801. The module comment now stops after the durable backend-selection statement; the version-specific dependency-resolution details have been removed. Formatting, both backend configurations, and Sequoia Clippy pass locally.

Minghan2005:
Addressed in 99a3751. The single PR commit is now based directly on current master at da389eb (v0.27.0), and the Sequoia changelog entry is under `## Unreleased` / `### Added`, not the 0.27.0 release section. Both complete backend configurations, doctests, no-feature checks, formatting, and strict Clippy pass on the rebased tree.

